PR #6, opened 2020-07-07
Fix installation instructions

jdreesen:
There's no version 2 yet. 
I removed the version completely to let composer choose the best matching version automatically.


drupol:
Thanks !

PR #40, opened 2020-11-05
New asyncMap() operation (featuring amphp!)

drupol:
This PR

* [x] Add a new operation: "`asyncMap(callable $callback)`"

Related to https://github.com/amphp/parallel-functions/issues/25

### DIFF
--- a/src/Operation/AsyncMap.php
+++ b/src/Operation/AsyncMap.php
@@ -1,0 +1,103 @@
+<?php
+
+declare(strict_types=1);
+
+namespace loophp\collection\Operation;
+
+use Amp\Emitter;
+use Amp\Promise;
+use Closure;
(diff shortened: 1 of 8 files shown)


src/Operation/AsyncMap.php
```diff
@@ -0,0 +1,103 @@
...
+use Generator;
+use Iterator;
+
+use function Amp\ParallelFunctions\parallel;
+use function Amp\Promise\all;
+use function Amp\Promise\wait;
+use function function_exists;
+
+// phpcs:disable
+if (false === function_exists('parallel')) {
```
kelunik, line 20: You need the full name here I guess?

drupol: Fixed, thanks !

PR #41, opened 2020-11-06
Fix `asyncMap` warning in the docs

jdreesen:
Shouldn't it say `non-deterministic` if the order cannot be guaranteed?


drupol (approved):
Oops, indeed, thanks ! I forgot it when rephrasing the sentence.

PR #49, opened 2020-12-08
Update ScanLeft so it doesn't go further if the iterator is not valid

### LINKED ISSUE #48: implode() would be better if it returned null/'' on empty array.
## Steps required to reproduce the problem

1. return Collection::fromIterable([])->implode();

## Expected Result
null or ''
* 

## Actual Result
[0 => '']
*

drupol:
This PR

* [x] Fixes #48 
* [x] Has a small behavior change that will be documented


drupol:
Merging this despite the unrelated CS issue (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4435#issuecomment-740780059).

PR #63, opened 2021-01-29
fix: Update behavior of Apply operation.

### LINKED ISSUE #62: apply() functionality
I'm not sure if this is a bug or if I just don't understand how this is supposed to work.

The documentation for Collection::apply() states:

> Execute a callback for each element of the collection without altering the collection item itself.
> 
> If the callback does not return true then it stops.

To me, this implies functionality similar to foreach. Returning `false` should stop the callback from executing for the items that follow.

In reality the apply() method executes the callback for every item regardless of the return value. However, if you supply several callbacks, returning `false` from one of them stops execution of the _following callbacks_, which is not the same.

After the following example, I would expect `$count === 2` but in reality `$count === 3`.

```php
$count = 0;

$fn = function($value) use (&$count) {
    $count++;
    if ($value === 'b') {
        return false;
    }

    return true;
}

Collection::fromIterable(['a', 'b', 'c'])
    ->apply($fn)
    ->all();
```

After running this code, with two callbacks (the same `$fn` from above), I would expect `$count === 2`, or possibly `$count === 4`, but in reality `$count === 5`:

```php
$count = 0;

Collection::fromIterable(['a', 'b', 'c'])
    ->apply($fn, $fn)
    ->all();
```

The callbacks will be called in this order:

'a': callback 1, callback 2
'b': callback 1,
'c': callback 1, callback 2

If this is the intended behaviour, is there another way to achieve foreach-like behaviour?

drupol:
This PR

* [x] Update behavior of Apply operation  
* [x] BREAKING CHANGE: yes
* [x] Add relevant tests (not yet done)
* [x] Update documentation

Fixes #62

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -190,10 +190,11 @@ Signature: ``Collection::append(...$items);``
 apply
 ~~~~~
 
-Execute a callback for each element of the collection without
-altering the collection item itself.
+Execute callback(s) on each element of the collection.
 
-If the callback does not return `true` then it stops.
+Iterates on the collection items regardless of the return value of the callback.
+
+If the callback does not return `true` then it stops applying callbacks on subsequent items.
 
 Interface: `Applyable`_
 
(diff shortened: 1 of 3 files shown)


drupol:
Ready for review @slackernrrd !

slackernrrd (approved):

PR #72, opened 2021-03-17
update psalm type for map operation

### LINKED ISSUE #71: why does map expect return type to be a subtype
```php
class Quantity {
  public float $value;
  public string $unitCode;
}
 // psalm error here
$qtys = $value = Collection::fromIterable($qtys)->map(fn (Quantity $qty): float => $qty->value);
```
The canonical type definition of map 
`map : (T -> V) -> List T -> List V`, 
where T and V are generic type variables, as far as i know there is no requirement of V being a sub type of T in other languages

aszenz:
Fixes #71 , by updating map type to follow the canonical type definition

### DIFF
--- a/src/Contract/Operation/Mapable.php
+++ b/src/Contract/Operation/Mapable.php
@@ -8,9 +8,9 @@ use Iterator;
 use loophp\collection\Contract\Collection;
 
 /**
- * @psalm-template TKey
  * @psalm-template TKey of array-key
  * @psalm-template T
+ * @psalm-template V
  */
 interface Mapable
 {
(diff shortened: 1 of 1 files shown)


drupol:
Thanks!

drupol:
I just noticed that we should have fixed the [implementation](https://github.com/loophp/collection/blob/master/src/Operation/Map.php) as well. It seems to be a more complex job! Care to give a try?

aszenz:
> I just noticed that we should have fixed the [implementation](https://github.com/loophp/collection/blob/master/src/Operation/Map.php) as well. It seems to be a more complex job! Care to give a try?

Sure, I'll give a try, is the fix related to the same issue or a different one

drupol:
Feel free to do what suits you best :) thanks !

aszenz:
@drupol 
I somehow kept wanting to do it, but couldn't, anyways i made a pr, not sure i fully understand how the map operation works for a collection of callable's so let me know if there's anything obliviously wrong with it

PR #69, opened 2021-02-25
Fix minimum PHP version in the docs

jdreesen:
This PR

* [x] Fixes the minimum PHP version in the docs (it's PHP >= 7.4 [since version 3](https://github.com/loophp/collection/releases/tag/3.0.0))

PR #73, opened 2021-03-18
Fix minimum PHP version in the readme

jdreesen:
This PR

* [x] Fixes the minimum PHP version in the readme (it's PHP >= 7.4 since version 3)

It is a followup of #69


drupol:
Thanks!

PR #74, opened 2021-04-02
collection: implements missing \Countable interface

Lctrs:
`loophp\collection\Collection` has a `count()` method, but `loophp\collection\Contract\Collection` doesn't implement the `\Countable` interface, making it useless. This PR fix this.

### DIFF
--- a/src/Contract/Collection.php
+++ b/src/Contract/Collection.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace loophp\collection\Contract;
 
+use Countable;
 use Iterator;
 use IteratorAggregate;
 use JsonSerializable;
@@ -229,6 +230,7 @@ interface Collection extends
     Combineable,
     Compactable,
     Containsable,
+    Countable,
     Currentable,
     Cycleable,
     Diffable,


drupol:
Hello,

Thanks for your PR!

About the changes, I was a bit reluctant to add this interface.

This example shows why.

```php
$fibonacci = static function (int $a = 0, int $b = 1): array {
    return [$b, $b + $a];
};

$c = Collection::unfold($fibonacci)
    ->count();

var_dump($c); // This will crash.
```

As this library can handle potential infinite stream of data, adding the `\Countable` interface was not making much sense, according to me.

Let me think a bit about it.

drupol:
Ok, I gave a second though. There are other things like that in loophp/collection, so, there is no reason why I should block this PR.

drupol:
Thanks for the contribution ! Hope loophp/collection is useful for you!

Lctrs:
> Ok, I gave a second though. There are other things like that in loophp/collection, so, there is no reason why I should block this PR.

Yeah, I had the same conclusion.



> Thanks for the contribution ! Hope loophp/collection is useful for you!

Thanks to you ! I love it.

drupol:
Excellent :-)

PR #77, opened 2021-05-08
Fix map type

aszenz:
Follows https://github.com/loophp/collection/pull/72
Related to https://github.com/loophp/collection/issues/71


drupol:
Hi!

Thanks for your contribution again.

I tried to make a commit but I didn't succeeded, in the end it created a new branch containing your commit and mine.
Anyway, here's the commit that I made: https://github.com/loophp/collection/commit/db24da6e89d4cbe5ce396f77cd5def7fa2c0b503

It fixes all the Psalm issues on the Map operation.

What do you think?

aszenz:
> Hi!
> 
> Thanks for your contribution again.
> 
> I tried to make a commit but I didn't succeeded, in the end it created a new branch containing your commit and mine.
> Anyway, here's the commit that I made: [db24da6](https://github.com/loophp/collection/commit/db24da6e89d4cbe5ce396f77cd5def7fa2c0b503)
> 
> It fixes all the Psalm issues on the Map operation.
> 
> What do you think?

The map function docs say:
> Apply one or more callbacks to a collection and use the return value.

I assume that map takes an array of callable's each of which transforms the collection one after the other.

After some thought about the actual function's behaviour, I think such a function can't be statically typed without losing it's behaviour.

Let's take a concrete example:
```php
// say i have a few data classes defined as
class Country {
  public function __construct(
    public string $name
  ) {}
}
class Address {
   public function __construct(
    public string $name,
    public Country $country
  ) {}
}
class Person {
  public function __construct(
    public string $name,
    public Address $address
  ) {}
}
// and some instances
$country1 = new Country('UK');
$address1 = new Address('foo', $country1);
$person1 = new Person('bar', $address1);

$country2 = new Country('FR');
$address2 = new Address('baz', $country2);
$person2 = new Person('kaz', $address2);

// this is the collection we intend to map(or transform)
$arr = [$person1, $person2];

// the map operation defined here is perfectly valid code and should work but psalm will complain still
$z = Map::of([
  fn(Person $p): Address => $p->address,  // type is Person -> Address
  fn(Address $a): Country => $a->country,  // type is Address -> Country
  fn(Country $c): string => $c->name // type is Country -> string
]);
\assert($z(Collection::from($arr)) === ['UK', 'FR']);
```

Each individual closure's input depends on the previous closure's output (or the input data in case of first closure element), this relationship cannot be statically typed as far I know. 

The commit https://github.com/loophp/collection/commit/db24da6e89d4cbe5ce396f77cd5def7fa2c0b503 assumes that each closure in the callbacks array has to be of the same type ( input `T|V`, output always `V`) which need not to be the case and is in fact limiting the function's behaviour as the above example shows.
Yes psalm would not complain but the function is no longer useful if we follow it's types.

The canonical map operation only takes one callback, which can be easily typed.
In this case because we take multiple callbacks, it becomes difficult to type.

Let me know if I have misunderstood the actual function's behaviour or if anything is unclear.

drupol:
You're absolutely right. This is how a `map` is working.

However, in `loophp/collection`, most arguments/parameters are `variadic`, so we have to type things in a different way.

If you think things should be different, let's continue the discussion.

aszenz:
> However, in loophp/collection, most arguments/parameters are variadic, so we have to type things in a different way.

That's fine, in that case the right type for the Map function is like `callable(mixed, array-key, Iterator<array-key, mixed>): mixed`, 

Using `mixed` means we lose static type safety but the function can be used as intended then. 
Right now the `map` function is not very useful for multiple callable's without getting complained by psalm.

Another option is to have two `map` functions once `variadic` and another `monodic`.

AlexandruGG:
@drupol what are your thoughts on the above? ☝️

drupol:
I also think the best solution is to get rid of the variadic `::map()`, when I did it at first, I though it was a good idea :disappointed: 

Even if I like the idea to have `::map1()` and `::mapN()`, I really want to keep it simple. Having multiple functions to do a `map` just to get a better typing score is not really what I want.
 
Also, changing the variadic function into non-variadic will be a BC break.

To be honest, I don't know how to deal with this issue.

AlexandruGG:
> I also think the best solution is to get rid of the variadic `::map()`, when I did it at first, I though it was a good idea 😞
> 
> Even if I like the idea to have `::map1()` and `::mapN()`, I really want to keep it simple. Having multiple functions to do a `map` just to get a better typing score is not really what I want.
> 
> Also, changing the variadic function into non-variadic will be a BC break.
> 
> To be honest, I don't know how to deal with this issue.

I don't think we need multiple `map1`, `map2`, ..., `mapN`, only two versions: `map` and `mapN`. I think this is still keeping it simple because they are meant for different purposes.
- `mapN` retains the functionality that map currently has, but with updated type hints
- `map` retains the same functionality for now, but triggers a deprecation notice _if multiple callbacks are passed to it_, with a message to recommend using `mapN` instead for that case. Also we can update the type hints to resemble the monadic version.

And then in the next major version (5.0 or 4.1, not sure how you want to consider) `map` is made to only accept one callback.

drupol:
I guess this can now be closed?

AlexandruGG:
> I guess this can now be closed?

@drupol I'm not sure, what's the conclusion? You didn't give your thoughts on this: https://github.com/loophp/collection/pull/77#issuecomment-847352407

drupol:
I guess this can be closed for multiple reasons:

1. We are no more using `TKey of array-key`
2. We have a strict typing in place thanks to #102
3. Regarding the need of having `::map` and `::mapN`, I don't close the door, but I think it's a bit out of scope of this PR. Maybe we should open a discussion?

AlexandruGG:
@drupol @aszenz this can be closed now with the merge of #112

PR #82, opened 2021-05-15
feat: Add Eager operation.

drupol:
This PR

* [x] Add an eager operation.
* [ ] Has documentation
* [ ] Has tests

PR #83, opened 2021-05-15
Feature: Add Squash operation

### LINKED ISSUE #80: FlatMap + Eager Usage
Hello,

Firstly, thank you for creating this library! I was also looking into writing a Collections library to replace the Doctrine Collections I typically use in Symfony projects because I couldn't find any good replacement with sufficient functionality, strict typing, proper code analysis and code quality checks etc. And just today I found your library, which looks awesome! 😄

I have two main questions:
1. I can't see any `flatMap` operation being available (also called `bind` in other languages). I know one can do `map`+ `flatten` but I think it's a common enough operation that having it would be nice. What do you think? I'd be happy to contribute!

2. If I want to apply a collection operation _eagerly_ rather than _lazily_ in a certain instance, but I still want to keep the collection rather than transform to an array, what's the best way to do it? Is it to call `apply()`? Let's say I'm calling `map()` with a function that throws an exception, and then returning the collection. I want the exception to be thrown at that level rather than at a level higher, where the collection is used. Due to the lazy nature, simply calling `map` will not cause the exception to be thrown.

I hope I've explained this well enough but let me know if I should clarify or provide an example! Thanks!

drupol:
This PR

* [x] Add Squash operation
* [x] Has relevant interfaces
* [x] Has tests
* [x] Has documentation

Follows #82.
Related to #80.
Fixes #80.


src/Operation/Squash.php
```diff
@@ -0,0 +1,72 @@
...
+                    static function (Iterator $iterator) use ($throwOnException): Closure {
+                        // As keys can be of any type in this library.We cannot use
+                        // iterator_to_array() because it expect keys to be int|string.
+                        try {
+                            foreach ($iterator as $key => $value) {
+                            }
+                        } catch (Throwable $e) {
+                            if (true === $throwOnException) {
+                                throw $e;
+                            }
```
AlexandruGG, line 49: what's the use case here for swallowing the exception if `throwOnException` is _false_? Just trying to understand what this could be used for.

drupol: This is something that came to my mind. We could imagine to ignore them or not. I don't like it so much, I'm still thinking about doing that in a different way. Do you have a better suggestion?

drupol: After thinking more about it, I removed it.

drupol:
Ready for another review.

src/Operation/Squash.php
AlexandruGG: Ignoring exceptions I think would be a bit unexpected. A different use case I can think of is if you passed a closure which you wanted to execute in the try-catch - let's say you wanted to add some logging if an exception occurs and then re-throw it. However, I feel like that's going more towards a method that you would want to use specifically for exception handling, so probably best to exclude it from this 👍

src/Collection.php
```diff
@@ -712,6 +712,11 @@ public function split(int $type = Operation\Splitable::BEFORE, callable ...$call
         return new self(Split::of()($type)(...$callbacks), $this->getIterator());
     }
 
+    public function squash(): CollectionInterface
+    {
+        return self::fromIterable($this->pack()->all())->unpack();
```
AlexandruGG, line 717: If you need I can help adding the docs for this method, but would just like to understand a bit better how this works. I'm looking at the docs for `pack` and `unpack` but still don't get why they're needed here 😓.

If you don't mind can you explain what's the difference between this and doing the below?

```PHP
return self::fromIterable($this->all());

AlexandruGG: I suspect it has to do with the collections allowing keys that are not only `int|string` as well as duplicate keys, right?

drupol: Yes exactly. `iterator_to_array()` will fail if keys are different from `int|string`.

See: https://3v4l.org/FW0Ca

drupol:
Ready for the last review!

AlexandruGG (approved):
Thank you for working on this so quickly! 🚀

PR #84, opened 2021-05-15
Update Documentation

AlexandruGG:
This PR updates the API documentation and README with details similar to other examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ This library has been inspired by:
 
 ## Usage
 
-Check out the [usage][32] and [examples][46] pages, they contains trivial and
+Check out the [usage][32] page for both trivial and
 more advanced use cases.
 
 ## Documentation
@@ -236,7 +236,6 @@ For more detailed changelogs, please check [the release changelogs][45].
 [33]: https://loophp-collection.rtfd.io
 [28]: https://loophp-collection.readthedocs.io/en/latest/pages/api.html
 [32]: https://loophp-collection.readthedocs.io/en/latest/pages/usage.html
-[46]: https://loophp-collection.readthedocs.io/en/latest/pages/examples.html
 [34]: https://github.com/loophp/collection/issues
 [2]: https://github.com/loophp/collection/actions
 [35]: http://www.phpspec.net/
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -158,8 +158,8 @@ html_static_path = ['_static']
 
 
 def setup(app):
-    app.add_stylesheet('custom.css')
-    app.add_stylesheet('highlight.css')
+    app.add_css_file('custom.css')
+    app.add_css_file('highlight.css')
 
 # Custom sidebar templates, maps document names to template names.
 #html_sidebars = {}
@@ -291,5 +291,3 @@ texinfo_documents = [
 
 # If true, do not generate a @detailmenu in the "Top" node's menu.
 #texinfo_no_detailmenu = False
-
-
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -455,7 +455,7 @@ Signature: ``Collection::contains(...$value);``
         ->contains('d'); // [false]
 
     if ($collection->contains('d')->current()) {
-	    // do something
+        // do something
     }
 
 current
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -158,6 +158,18 @@ Interface: `Allable`_
 
 Signature: ``Collection::all();``
 
+.. code-block:: php
+
+        $generator = static function (): Generator {
+            yield 0 => 'a';
+            yield 1 => 'b';
+            yield 0 => 'a';
+            yield 2 => 'c';
+        };
+
+        Collection::fromIterable($generator())
+            ->all(); // [0 => 'a', 1 => 'b', 2 => 'c']
+
 append
 ~~~~~~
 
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -236,7 +236,6 @@ For more detailed changelogs, please check [the release changelogs][45].
 [33]: https://loophp-collection.rtfd.io
 [28]: https://loophp-collection.readthedocs.io/en/latest/pages/api.html
 [32]: https://loophp-collection.readthedocs.io/en/latest/pages/usage.html
-[46]: https://loophp-collection.readthedocs.io/en/latest/pages/examples.html
```
AlexandruGG, line 239: This is a dead URL, I'm not sure if it pointed to a page that no longer exists or the page exists but it was just renamed, please advise.

docs/conf.py
```diff
@@ -158,8 +158,8 @@
 
 
 def setup(app):
-    app.add_stylesheet('custom.css')
-    app.add_stylesheet('highlight.css')
```
AlexandruGG, line 162: when trying to preview the `.rst` files in VSCode it was failing with this error: https://github.com/sphinx-doc/sphinx/issues/7747. The change I made fixes it for me, is it ok to make? I'm not too familiar with Sphinx usage

docs/pages/api.rst
```diff
@@ -435,6 +447,11 @@ Interface: `Containsable`_
 
 Signature: ``Collection::contains(...$value);``
 
+.. code-block:: php
+
+    $collection = Collection::fromIterable(range('a', 'c'))
+        ->contains('d'); // [false]
```
AlexandruGG, line 453: this behaviour is based on the spec but I have to say it surprised me. Shouldn't the return type be just `bool`? I would expect to use this method as such:

```PHP

$collection = Collection::fromIterable([1, 2, 3]);

if ($collection->contains(2)) {
    // do something
}
```

How is the current implementation meant to be used?

docs/pages/api.rst
```diff
@@ -454,16 +471,16 @@ Signature: ``Collection::current(int $index = 0);``
 cycle
 ~~~~~
 
-Cycle around a collection of items.
+Cycle indefinitely around a collection of items.
 
 Interface: `Cycleable`_
 
-Signature: ``Collection::cycle(int $length = 0);``
```
AlexandruGG, line 461: I'm not sure if this was for an older version, but the current `cycle` method doesn't take a parameter and cycles indefinitely from what I can see

README.md
drupol: Thanks!

docs/pages/api.rst
drupol: Indeed, this was a choice that I made in order to make sure that the collection would always return a collection.
The 3 only methods that are not potentially returning a Collection are: `all`, `key`, `current`.
In this example, to get the boolean, you have to do:

```php
$collection = Collection::fromIterable([1, 2, 3]);

if ($collection->contains(2)->current()) {
    // do something
}
```

docs/pages/api.rst
drupol: Seems like this was from an old implementation. Maybe we should keep the signature, but just updating it:

` Signature: ``Collection::cycle();`` `

drupol (approved):
Ignore this approval comment, I hit the wrong button!

drupol (changes requested):
Thanks!

I made a review. All those changes are great!

There's a small things to update, as soon as it's ok, I merge this.

docs/pages/api.rst
```diff
@@ -158,6 +158,18 @@ Interface: `Allable`_
 
 Signature: ``Collection::all();``
 
+.. code-block:: php
+
+        $generator = static function (): Generator {
+            yield 0 => 'a';
+            yield 1 => 'b';
+            yield 0 => 'a';
```
drupol, line 166:
```suggestion
            yield 0 => 'c';
```

docs/pages/api.rst
```diff
@@ -158,6 +158,18 @@ Interface: `Allable`_
...
+            yield 2 => 'c';
```
drupol, line 167:
```suggestion
            yield 2 => 'd';
```

docs/pages/api.rst
```diff
@@ -158,6 +158,18 @@ Interface: `Allable`_
...
+        };
+
+        Collection::fromIterable($generator())
+            ->all(); // [0 => 'a', 1 => 'b', 2 => 'c']
```
drupol, line 171:
```suggestion
            ->all(); // [0 => 'c', 1 => 'b', 2 => 'd']
```

drupol (changes requested):

docs/pages/api.rst
AlexandruGG: that makes sense, I added a note [here](https://github.com/loophp/collection/pull/84/commits/21bdbe8bf781d00c5078dcde232e1ae3c48b65e7#diff-89f6324da13155a697ae0b42136e03a40892dd4edc4257ea1e451f0ea2cc3f80R148-R150) to highlight this a bit better and added to the [contains example](https://github.com/loophp/collection/pull/84/commits/21bdbe8bf781d00c5078dcde232e1ae3c48b65e7#diff-89f6324da13155a697ae0b42136e03a40892dd4edc4257ea1e451f0ea2cc3f80R457-R459)

drupol (approved):
Thanks for taking the time to do this, I'm very grateful !

AlexandruGG:
> Thanks!
> 
> I made a review. All those changes are great!
> 
> There's a small things to update, as soon as it's ok, I merge this.

Thank you, agreed the suggested change to that example makes it clearer 👍

AlexandruGG:
> Thanks for taking the time to do this, I'm very grateful !

I appreciate all the work that you've already put in this project! 💯

drupol:
Thanks again :)

Issue #56, opened 2020-12-27
collection not working with php 8.0


MartkCz:
PHP 8.0 introduces new expression named match, which is BC break for classes and functions named same as keyword "match"

https://github.com/loophp/collection/blob/master/src/Operation/Match.php

drupol:
Oh... This is unfortunate :(

I will rename it in v4, will work on it tonight.

Sorry for the inconveniences ...

drupol:
I just released the version 3.3.5 that address this issue.

Could you please check? And close the ticket if it's ok?

Also, I haven't the chance to work with PHP 8 yet, looks like you do, how it's behaving in PHP 8? Is it ok? Do you see a difference with PHP 7 ?

MartkCz:
I run php linter for all files and no syntax errors. Thank you!

drupol:
Great, thanks!

PR #86, opened 2021-05-16
Update API Documentation II

AlexandruGG:
Adding more examples and some clarifications to the API documentation.

I'm aiming to go through the whole document in this PR so I will keep it as draft until I've done so: 

* [x] Check methods down to 'h'
* [x] Check methods 'h' -> 'o'
* [x] Check methods 'p' -> 'z'

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,8 +1,8 @@
 # CONTRIBUTING
 
-We're using [Travis CI](https://travis-ci.com) as a continuous integration system.
+We're using [GitHub Actions](https://docs.github.com/en/actions) as a continuous integration system.
  
-For details, see [`.travis.yml`](../.travis.yml). 
+For details, see the workflows in [.github/workflows](../.github/workflows). 
  
 ## Tests
 
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -145,16 +145,15 @@ Another example
 Methods (operations)
 --------------------
 
-.. note::
-	Operations always returns a new collection object, with the exception of ``all``, ``count``, ``current``, ``key``.
-
+.. note:: Operations always returns a new collection object, with the exception of ``all``, ``count``, ``current``, ``key``.
 
 all
 ~~~
 
 Convert the collection into an array.
 
-This is a lossy operation because PHP array keys cannot be duplicated and must either be int or string.
+.. warning:: This is a lossy operation because PHP array keys cannot be duplicated and must either be int or string.
+            If you want to ensure no data is lost in the case of duplicate keys, look at the ``Collection::normalize()`` operation.
 
 Interface: `Allable`_
 
@@ -663,7 +662,7 @@ Signature: ``Collection::explode(...$items);``
 falsy
 ~~~~~
 
-Check if the collection contains falsy values.
+Check if the collection contains *any falsy* values. A value is determined to be *falsy* by applying a ``bool`` cast.
 
 Interface: `Falsyable`_
 
@@ -988,7 +987,7 @@ Signature: ``Collection::head();``
 ifThenElse
 ~~~~~~~~~~
 
-Execute a callback when a condition is met.
+Execute a callback when a condition is met. If no ``else`` callback is provided, the identity function is applied (elements are not modified).
 
 Interface: `IfThenElseable`_
 
@@ -1019,13 +1018,18 @@ Signature: ``Collection::ifThenElse(callable $condition, callable $then, ?callab
 implode
 ~~~~~~~
 
-Convert all the elements of the collection to a single string.
+Join all the elements of the collection into a single string using a glue provided or the empty string as default.
 
-The glue character can be provided, default is the empty character.
+Note internally ``foldLeft`` is used, which is why the result will have the last element's key.
 
 Interface: `Implodeable`_
 
 Signature: ``Collection::implode(string $glue = '');``
+
+.. code-block:: php
+
+    Collection::fromIterable(range('a', 'c'))
+        ->implode('-'); // [2 => 'a-b-c']
 
 init
 ~~~~
@@ -1094,8 +1098,12 @@ Signature: ``Collection::intersperse($element, int $every = 1, int $startAt = 0)
 
 .. code-block:: php
 
-    $collection = Collection::fromIterable(range('a', 'z'))
-        ->intersperse('foo', 3);
+    $collection = Collection::fromIterable(range('a', 'c'))
+        ->intersperse('x');
+    
+    foreach($collection as $item) {
+        var_dump($item); // 'x', 'a', 'x', 'b', 'x', 'c'
+    }
 
 key
 ~~~
@@ -1124,8 +1132,8 @@ Signature: ``Collection::keys();``
 
 .. code-block:: php
 
-    $collection = Collection::fromIterable(range('a', 'z'))
-        ->keys();
+    $collection = Collection::fromIterable(range('a', 'd'))
+        ->keys(); // [0, 1, 2, 3]
 
 last
 ~~~~
@@ -1161,12 +1169,8 @@ Signature: ``Collection::limit(int $limit);``
 
 .. code-block:: php
 
-    $fibonacci = static function ($a = 0, $b = 1): array {
-        return [$b, $a + $b];
-    };
-
-    $collection = Collection::unfold($fibonacci)
-        ->limit(10);
+    $even = Collection::range(0, \INF, 2)
+        ->limit(5); // [0, 2, 4, 6, 8]
 
 lines
 ~~~~~
@@ -1180,20 +1184,20 @@ Signature: ``Collection::lines();``
 .. code-block:: php
 
     $string = <<<'EOF'
-    The quick brow fox jumps over the lazy dog.
+    The quick brown fox jumps over the lazy dog.
 
     This is another sentence.
     EOF;
 
     Collection::fromString($string)
-        ->lines();
+        ->lines(); // ['The quick brown fox jumps over the lazy dog.', '', 'This is another sentence.']
 
 map
 ~~~
 
 Apply one or more supplied callbacks to every item of a collection and use the return value.
 
-.. warning:: Keys are preserved, use the "normalize" operation if you want to re-index the keys.
+.. warning:: Keys are preserved, use the ``Collection::normalize`` operation if you want to re-index the keys.
 
 Interface: `Mapable`_
 
@@ -1205,8 +1209,8 @@ Signature: ``Collection::map(callable ...$callbacks);``
         return $value * 2;
     };
 
-    $collection = Collection::fromIterable(range(1, 100))
-        ->map($mapper);
+    $collection = Collection::fromIterable(range(1, 5))
+        ->map($mapper); // [2, 4, 6, 8, 10]
 
 match
 ~~~~~
@@ -1234,30 +1238,38 @@ Signature: ``Collection::match(callable $callback, ?callable $matcher = null);``
 merge
 ~~~~~
 
-Merge one or more collection of items onto a collection.
+Merge one or more iterables onto a collection.
 
 Interface: `Mergeable`_
 
-Signature: ``Collection::merge(...$sources);``
+Signature: ``Collection::merge(iterable ...$sources);``
 
 .. code-block:: php
 
-    $collection = Collection::fromIterable(range(1, 10))
-        ->merge(['a', 'b', 'c'])
+    Collection::fromIterable(range(1, 5))
+        ->merge(range(6, 10)) // range(1, 10)
+
+    $collection = Collection::fromIterable(['a', 'b', 'c'])
+        ->merge(Collection::fromIterable(['d', 'e']);
+    
+    $collection->all(); // ['d', 'e', 'c'] -> 'a' and 'b' are lost due to key overlap
+    $collection->normalize()->all() // ['a', 'b', 'c', 'd', 'e']
 
 normalize
 ~~~~~~~~~
 
 Replace, reorder and use numeric keys on a collection.
 
+.. note:: If you want to retrieve collection elements as an array via ``Collection::all()`` instead of
+        consuming the collection through a ``foreach``, most often you will want to use this method
+        before the array transformation in order to prevent data loss.
+
 Interface: `Normalizeable`_
 
 Signature: ``Collection::normalize();``
 
-.. code-block:: php
-
-    $collection = Collection::fromIterable(['a' => 'a', 'b' => 'b', 'c' => 'c'])
-        ->normalize();
+.. literalinclude:: code/operations/normalize.php
+  :language: php
 
 nth
 ~~~
@@ -1270,13 +1282,13 @@ Signature: ``Collection::nth(int $step, int $offset = 0);``
 
 .. code-block:: php
 
-    $collection = Collection::fromIterable(range(10, 100))
-        ->nth(3);
+    $collection = Collection::fromIterable(range(1, 20))
+        ->nth(5); // [0 => 1, 5 => 6, 10 => 11, 15 => 16]
 
 nullsy
 ~~~~~~
 
-Check if the collection contains *nullsy* values.
+Check if the collection contains *only nullsy* values.
 
 *Nullsy* values are:
 
@@ -1290,19 +1302,17 @@ Interface: `Nullsyable`_
 
 Signature: ``Collection::nullsy();``
 
-only
-~~~~
-
-Get items having corresponding given keys.
-
-Interface: `Onlyable`_
-
-Signature: ``Collection::only(...$keys);``
-
 .. code-block:: php
 
-    $collection = Collection::fromIterable(range(10, 100))
-        ->only(3, 10, 'a', 9);
+    $nullsy = Collection::fromIterable([null, null])
+        ->nullsy(); // [true]
+
+    $nonNullsy = Collection::fromIterable(['a', null, 'c'])
+        ->nullsy(); // [false]
+
+    if ($falsyCollection->nullsy()->current()) {
+        // do something
+    }
 
 pack
 ~~~~
@@ -2308,7 +2318,6 @@ Signature: ``Collection::zip(iterable ...$iterables);``
 .. _Normalizeable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Normalizeable.php
 .. _Nthable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Nthable.php
 .. _Nullsyable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Nullsyable.php
-.. _Onlyable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Onlyable.php
 .. _Packable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Packable.php
 .. _Padable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Padable.php
 .. _Pairable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Pairable.php
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -152,7 +152,7 @@ all
 
 Convert the collection into an array.
 
-.. warning:: This is a lossy operation because PHP array keys cannot be duplicated and must either be int or string.
+.. warning:: This is a lossy operation because PHP array keys cannot be duplicated and must either be ``int`` or ``string``.
             If you want to ensure no data is lost in the case of duplicate keys, look at the ``Collection::normalize()`` operation.
 
 Interface: `Allable`_
@@ -465,7 +465,7 @@ Signature: ``Collection::contains(...$value);``
 current
 ~~~~~~~
 
-Get the value of an item in the collection given a numeric index or the default 0.
+Get the value of an item in the collection given a numeric index or the default ``0``.
 
 Interface: `Currentable`_
 
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -1537,11 +1537,13 @@ Signature: ``Collection::reverse();``
 rsample
 ~~~~~~~
 
-Work in progress... sorry.
+Take a random sample of elements of items from a collection. Accepts a probability parameter
+which will influence the number of items sampled - higher probabilities increase the chance of
+sampling close to the entire collection.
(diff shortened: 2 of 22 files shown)


.github/CONTRIBUTING.md
```diff
@@ -1,8 +1,8 @@
 # CONTRIBUTING
 
-We're using [Travis CI](https://travis-ci.com) as a continuous integration system.
+We're using [GitHub Actions](https://docs.github.com/en/actions) as a continuous integration system.
  
-For details, see [`.travis.yml`](../.travis.yml). 
+For details, see the workflows in [.github/workflows](../.github/workflows).
```
AlexandruGG, line 5: related to `GrumPHP` when I try to run it on my machine it's failing in `master` (before any code changes). 

Is this something you're aware of @drupol or am I missing some dependency? I already installed composer dependencies.

```
GrumPHP is sniffing your code!

Running tasks with priority 0!
==============================

Running task 1/14: license... ✔
Running task 2/14: composer_require_checker... ✘
Running task 3/14: composer... ✔
Running task 4/14: composer_normalize... ✔
Running task 5/14: yamllint... ✔
Running task 6/14: jsonlint... ✔
Running task 7/14: phplint... ✔
Running task 8/14: twigcs...
Running task 9/14: phpcsfixer... ✔
Running task 10/14: phpcs... ✔
Running task 11/14: phpstan... ✘
Running task 12/14: psalm... ✔
Running task 13/14: phpspec... ✘
Running task 14/14: infection... ✘

File: docs/pages/api.rst [hunk 2]
```diff
@@ -348,15 +348,19 @@ collapse
 
 Collapse a collection of items into a simple flat collection.
 
+.. warning:: Key differences compared to ``flatten`` are that only one level will be collapsed and values at the bottom level will be removed.
```
AlexandruGG, line 351: I looked into the tests to figure this out, please let me know if there are any other differences which should be mentioned

.github/CONTRIBUTING.md
drupol: Strange.

In order to help, could you run:

`php -v`

Then:

`./vendor/bin/infection run -vvv`

docs/pages/api.rst
drupol: That's the proper behavior indeed.

.github/CONTRIBUTING.md
AlexandruGG: I'm running on PHP 8 maybe that's why? 😅

```php -v
PHP 8.0.3 (cli) (built: Mar  4 2021 20:45:17) ( NTS )
Copyright (c) The PHP Group
Zend Engine v4.0.3, Copyright (c) Zend Technologies
    with Zend OPcache v8.0.3, Copyright (c), by Zend Technologies
```

Infection fails because PHPSpec fails:

```
vendor/bin/phpspec run                                                                       1 ↵

      loophp\collection\Collection

  31  ✔ can append
  64  ✔ can apply
 182  ✔ can associate
 215  ! can asyncMap (109ms)
        exception [err:Error("Loop exceptionally stopped without resolving the promise")] has been thrown.
 232  ✔ can be constructed from a file
 242  ✔ can be constructed from a generator
 261  ✘ can be constructed from a stream
        expected exception of class "InvalidArgumentException", but got [obj:TypeError] with the
        message: "get_resource_type(): Argument #1 ($resource) must be of type resource, GdImage given"
 282  ✔ can be constructed from a string
 295  ✔ can be constructed from an iterable
 320  ✔ can be constructed from an iterator
 327  ✔ can be constructed from empty
 336  ✔ can be constructed with a closure
 383  ✔ can be constructed with an arrayObject
 391  ✔ can be instantiated with withClosure
 409  ✔ can be json encoded
 423  ✔ can be returned as an array
 429  ✔ can cache
 465  ✔ can chunk
 484  ✔ can coalesce
 503  ✔ can collapse
 531  ✔ can column
 568  ✔ can combinate
 620  ✔ can combine
 632  ✔ can compact
 645  ✔ can contains
 664  ✔ can convert use a string as parameter
 673  ✔ can count its items
 680  ✔ can current
 700  ✔ can cycle
 724  ✔ can diff
 735  ✔ can diffKeys
 748  ✔ can distinct
 757  ✔ can do the cartesian product
 776  ✔ can drop
 787  ✔ can dropWhile
string(6) "------"
array(3) {
  ["name"]=>
  string(5) "debug"
  ["key"]=>
  int(0)
  ["value"]=>
  string(1) "a"
}
array(3) {
  ["name"]=>
  string(5) "debug"
  ["key"]=>
  int(1)
  ["value"]=>
  string(1) "b"
}
string(6) "------"
array(3) {
  ["name"]=>
  string(5) "debug"
  ["key"]=>
  int(0)
  ["value"]=>
  string(1) "a"
}
array(3) {
  ["name"]=>
  string(5) "debug"
  ["key"]=>
  int(1)
  ["value"]=>
  string(1) "b"
}
string(6) "------"
 827  ✔ can dump
 908  ✔ can duplicate
 921  ✔ can every
 977  ✔ can explode
1026  ✔ can falsy
1041  ✔ can filter
1075  ✔ can flatten
1122  ✔ can flip
1135  ✔ can fold from the left
1149  ✔ can fold from the right
1163  ✔ can foldleft1
1178  ✔ can foldright1
1193  ✔ can forget
1200  ✔ can frequency
1227  ✔ can get
1238  ✔ can get an iterator
1247  ✔ can get current
1264  ✔ can get key
1281  ✔ can get the first item
1292  ✔ can get the last item
1320  ✔ can group
1348  ✔ can groupBy
1425  ✔ can has
1486  ✔ can head
1495  ✔ can if then else
1524  ✔ can implode
1535  ✔ can init
1562  ✔ can inits
1574  ✔ can intersect
1585  ✔ can intersectKeys
1606  ✔ can intersperse
1721  ✔ can key
1737  ✔ can keys
1744  ✔ can limit
1761  ✔ can lines
1780  ✔ can map
1809  ✔ can match
1837  ✔ can merge
1870  ✔ can normalize
1893  ✔ can nth
1904  ✔ can nullsy
1923  ✔ can pack
1938  ✔ can pad
1947  ✔ can pair
1990  ✔ can partition
2021  ✔ can permutate
2061  ✔ can pipe
2101  ✔ can pluck
2178  ✔ can prepend
2199  ✔ can random
2237  ✔ can reduction
2249  ✔ can reverse
2264  ✔ can rsample
2275  ✔ can scale
2291  ✔ can scanleft
2316  ✔ can scanleft1
2331  ✔ can scanright
2358  ✔ can scanright1
2383  ✔ can shuffle
2427  ✔ can since
2499  ✔ can slice
2514  ✔ can sort
2599  ✔ can span
2621  ✔ can split
2649  ✔ can squash
2667  ✔ can tail
2674  ✔ can tails
2688  ✔ can takeWhile
2726  ✔ can transpose
2777  ✔ can truthy
2796  ✔ can unfold
2841  ✔ can unlines
2862  ✔ can unpack
2899  ✔ can unpair
2918  ✔ can until
2955  ✔ can unwindow
3090  ✔ can unwords
3118  ✔ can unwrap
3138  ✔ can unzip
3156  ✔ can use range
3205  ✔ can use range with value 1
3215  ✔ can use times with a callback
3228  ✔ can use times without a callback
3246  ✔ can window
3381  ✔ can words
3410  ✔ can wrap
3425  ✔ can zip
3457  ✔ fix bug 57
3482  ✔ is initializable

      loophp\collection\Iterator\ArrayCacheIterator

  19  ✔ can cache an iterator of type generator
  49  ✔ is initializable

      loophp\collection\Iterator\IterableIterator

  17  ✔ can be constructed with a callable which is not a generator
  23  ✔ can get a key
  33  ✔ can rewind
  59  ✔ can use next
  73  ✔ is initializable

      loophp\collection\Iterator\PsrCacheIterator

  21  ✔ can cache data
  83  ✔ can get the inner iterator
  93  ✔ is initializable

      loophp\collection\Iterator\RandomIterator

  19  ✔ can build an iterator with a random seed
  70  ✔ can build an iterator without a random seed
  82  ✔ can get the innerIterator
  87  ✔ can rewind
 125  ✔ is initializable

      loophp\collection\Operation\Coalesce

  18  ✔ can coalesce
  49  ✔ is initializable

      loophp\collection\Operation\Current

  18  ✔ can get current
  33  ✔ is initializable

      loophp\collection\Operation\Limit

  18  ✔ can set a limit
  33  ✔ can set an offset
  48  ✔ does not change anything when no parameters are provided
  59  ✔ is initializable

----  failed examples

        loophp/collection/Collection
 261  ✘ can be constructed from a stream
        expected exception of class "InvalidArgumentException", but got [obj:TypeError] with the
        message: "get_resource_type(): Argument #1 ($resource) must be of type resource, GdImage given"

----  broken examples

        loophp/collection/Collection
 215  ! can asyncMap (109ms)
        exception [err:Error("Loop exceptionally stopped without resolving the promise")] has been thrown.


8 specs
147 examples (145 passed, 1 failed, 1 broken)
689ms
```

drupol: ah yes,... it's not yet PHP 8 fully compatible :(

docs/pages/api.rst
```diff
@@ -2254,7 +2320,6 @@ Signature: ``Collection::zip(iterable ...$iterables);``
 .. _Normalizeable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Normalizeable.php
 .. _Nthable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Nthable.php
 .. _Nullsyable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Nullsyable.php
-.. _Onlyable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Onlyable.php
```
AlexandruGG, line 2257: this no longer exists, was replaced by `IntersectKeys`

drupol: Yep, nicely spotted :)

docs/pages/api.rst
```diff
@@ -145,16 +145,15 @@ Another example
...
-
+.. note:: Operations always returns a new collection object, with the exception of ``all``, ``count``, ``current``, ``key``.
 
 all
 ~~~
 
 Convert the collection into an array.
 
-This is a lossy operation because PHP array keys cannot be duplicated and must either be int or string.
+.. warning:: This is a lossy operation because PHP array keys cannot be duplicated and must either be int or string.
```
drupol, line 155:
```suggestion
.. warning:: This is a lossy operation because PHP array keys cannot be duplicated and must either be ``int`` or ``string``.
```

docs/pages/api.rst
```diff
@@ -461,7 +465,7 @@ Signature: ``Collection::contains(...$value);``
 current
 ~~~~~~~
 
-Get the value of an item in the collection given a numeric index, default index is 0.
+Get the value of an item in the collection given a numeric index or the default 0.
```
drupol, line 468:
```suggestion
Get the value of an item in the collection given a numeric index or the default ``0``.
```

docs/pages/api.rst
```diff
@@ -965,14 +1019,19 @@ Signature: ``Collection::ifThenElse(callable $condition, callable $then, ?callab
 implode
 ~~~~~~~
 
-Convert all the elements of the collection to a single string.
+Join all the elements of the collection into a single string using a glue provided or the empty string as default.
 
-The glue character can be provided, default is the empty character.
+Note internally ``foldLeft`` is used, which is why the result will have the last element's key.
```
drupol, line 1024: Nice! Should we add a tips box for this?

.github/CONTRIBUTING.md
AlexandruGG: Hmm I thought that with https://github.com/loophp/collection/issues/56 sorted it would be.

Do you mean it might not be usable in other projects using PHP 8, or just that internally it's not compatible?

Also, is there a list of issues or milestones to hit in order to reach compatibility?

docs/pages/api.rst
AlexandruGG: Yeah that sounds good, will add 👍

.github/CONTRIBUTING.md
drupol: It will be compatible with PHP 8 in June because it will be my main focus during that month.

AlexandruGG: Sounds good, as I'm looking to use this in a project using PHP 8.

Could I help address some of the issues?

drupol: Of course ! PRs are very welcome and as you can see, I'm quite available :)

docs/pages/api.rst
```diff
@@ -1021,7 +1021,7 @@ implode
 
 Join all the elements of the collection into a single string using a glue provided or the empty string as default.
 
-Note internally ``foldLeft`` is used, which is why the result will have the last element's key.
+.. tip:: Internally this operation uses ``foldLeft``, which is why the result will have the last element's key.
```
drupol, line 1024: :+1:

.github/CONTRIBUTING.md
AlexandruGG: Cool, I will start looking into the issues I experienced on my local

docs/pages/api.rst
```diff
@@ -1493,13 +1532,23 @@ Signature: ``Collection::reverse();``
...
 
+Interface: `RSampleable`_
+
+Signature: ``Collection::rsample();``
+
+.. code-block:: php
+
+    $collection = Collection::fromIterable(range(1, 5));
+    $collection->rsample(1.0); // [1, 2, 3, 4, 5]
+    $collection->rsample(0.5); // will get about half of the elements at random
```
AlexandruGG, line 1550: I saw the description of `rsample` only says "work in progress" currently. What is this method based on? Is there anything we can link to? Is there a specific use case for it?

I tried to understand how it works and it looks to me like it returns a random set of elements based on the probability you supply: the higher the probability the more likely you will get close to all elements in the collection.

docs/pages/api.rst
```diff
@@ -1512,10 +1561,10 @@ Signature: ``Collection::scale(float $lowerBound, float $upperBound, ?float $wan
 .. code-block:: php
 
     $collection = Collection::range(0, 10, 2)
-        ->scale(0, 10);
+        ->scale(0, 10); // [0, 0.2, 0.4, 0.6, 0.8]
```
AlexandruGG, line 1564: `scale` is another one that I'd like to improve the description of, as it's quite hard to understand what exactly it does and where it could be used. I still don't fully get it 😅

docs/pages/api.rst
drupol: This is it exactly. Nothing more, nothing less.

docs/pages/api.rst
drupol: This is a very useful function.

Let's say that you have these numbers: `[0, 5,10,15,20]`

You would like to "scale" them using a another linear scale, let's say, from 0 to 12.

```php
$collection = Collection::fromIterable([0, 5, 10, 15, 20])
  ->scale(0, 20, 0, 12); // [0, 3, 6, 9, 12]
```

You can also get a logarithmic scale by adding an extra `base` parameter

```php
$collection = Collection::fromIterable([0, 5, 10, 15, 20])
  ->scale(0, 20, 0, 12, \M_E); // [1, 6.909, 9.454, 10.943, 12]
```

Let me know if you need more explanation.

docs/pages/api.rst
AlexandruGG: Alright nice, I will add that in

docs/pages/api.rst
AlexandruGG: gotcha, it has to do with [normalization](https://en.wikipedia.org/wiki/Normalization_(statistics)) and basically allows you to keep the same relative distance between values, but have different lower and upper bounds for the set of values

drupol: That's indeed exactly it.

docs/pages/api.rst
```diff
@@ -2133,16 +2062,17 @@ Signature: ``Collection::window(int $size);``
 
 .. code-block:: php
 
-     $data = range('a', 'z');
+     $data = range('a', 'd');
 
      Collection::fromIterable($data)
-        ->window(2)
-        ->all(); // [ ['a'], ['a', 'b'], ['b', 'c'], ['c', 'd'], ... ]
+        ->window(2); // [['a'], ['a', 'b'], ['a', 'b', 'c'], ['b', 'c', 'd']]
```
AlexandruGG, line 2068: This is the actual output of the method, as also exemplified by the [spec](https://github.com/loophp/collection/blob/master/spec/loophp/collection/CollectionSpec.php#L3223-L3239). 

However, this is really surprising to me. I would expect the following, given that the parameter passed is size of the window (2 in this case).

```PHP
Collection::fromIterable(range('a', 'd'))
    ->window(2); // [['a', 'b'], ['b', 'c'], ['c', 'd']]
```

Removing [this](https://github.com/loophp/collection/blob/master/src/Operation/Window.php#L45) gets it closer to that: `[['a'], ['a', 'b'] , ['b', 'c'], ['c', 'd']`. But I'm still not sure about the first element.

@drupol is this a bug or am I missing something?

docs/pages/api.rst
```diff
@@ -1853,7 +1799,8 @@ Signature: ``Collection::tail();``
 tails
 ~~~~~
 
-Returns the list of initial segments of its argument list, shortest last.
+Returns the list of initial segments of the collection, shortest last.
+Similar to applying ``tail`` successively and collecting all results in one list.
```
AlexandruGG, line 1803: once I read your blog and realised the Haskell influence many of these methods started to make more sense 😄 . The initial definition was pretty much like the Haskell one hehe, but in our case `tails` doesn't take arguments

docs/pages/api.rst
```diff
@@ -1355,42 +1418,36 @@ Signature: ``Collection::partition(callable ...$callbacks);``
...
 prepend
 ~~~~~~~
 
 Push an item onto the beginning of the collection.
 
-.. warning:: If prepended values overwrite existing values, you might find that this operation doesn't work correctly
+.. warning:: If prepended values overwrite existing values or keys, you might find that this operation doesn't work correctly
              when the collection is converted into an array.
              It's always better to never convert the collection to an array and use it in a loop.
              However, if for some reason, you absolutely need to convert it into an array, then use the
```
AlexandruGG, line 1453: I realised there are multiple similar warnings like this. I think it would be useful to have an initial section describing how to work with lazyness and with this library to avoid unexpected behaviour, loss of data etc. 

Perhaps in a subsequent PR I will put something together. That section needs to come before the API methods because the principles apply to pretty much every method.

docs/pages/api.rst
drupol: This is not a bug, basically the `Window` operation "adds" an amount of item, on top of the current one.

This is why if you set it 2, you might end up getting windows of 3 items. (_2 + 1 current_).

docs/pages/api.rst
drupol: Ahaha you went down into the rabbit hole :) congrats! Haskell is so beautiful that I wanted to provide the same kind in PHP.

However, `[tails](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-List.html#v:tails)` in Haskell doesn't take arguments.

docs/pages/api.rst
drupol: Yes, good idea. I already wrote about it here: https://loophp-collection.readthedocs.io/en/latest/pages/usage.html#manipulate-keys-and-values

It could be a starting point?

docs/pages/api.rst
AlexandruGG: Hmm I see, I will update the explanation to reflect that. Although the behaviour is still a bit unexpected, because a typical sliding window would only result in groupings of the same size. What's the reason for also having the first element separate?

docs/pages/api.rst
AlexandruGG: I like it, though my experience with it is pretty much limited to reading this [book](http://learnyouahaskell.com/) and trying out some exercises 😄

docs/pages/api.rst
AlexandruGG: Ah nice, I didn't see that section because I didn't go into all the Advanced usages. Can definitely be a starting point 👍

docs/pages/api.rst
drupol: Because it would add opinionated behavior and I'm trying to avoid that. Some people would need them, some would not.

However, if someone doesn't want to see the first elements which are not yet of the size of the sliding window, they can drop them.

```php
$input = array_combine(range('a', 'z'), range('a', 'z'));

$collection = Collection::fromIterable($input);

$output = $collection
    ->window(3)
    ->drop(3);

foreach ($output as $key => $value) {
    var_dump($value); 
}
```

Does it help?

docs/pages/api.rst
drupol: Same here and also this one, which is actually better: https://www.manning.com/books/get-programming-with-haskell

docs/pages/api.rst
AlexandruGG: Ok I think I get what you mean. 

The behaviour is as follows: `window(int $size)` starts building a window of elements until it gets to the size `1 + $size`, and then keeps moving along the elements of the collection keeping that same size. 

The behaviour I was confused about with the smaller size elements at first makes sense in this context because the window needs to first reach that given size.

docs/pages/api.rst
AlexandruGG: thanks, I'll check it out!

docs/pages/api.rst
drupol: That's precisely that.

drupol (approved):
What a work !!! Thanks !!!!!!!!!

AlexandruGG:
> What a work !!! Thanks !!!!!!!!!

Thank you! It's been fun to dive into the API of this. Stay tuned for more! 😁 

Btw [this syntactic sugar](https://github.com/loophp/collection/blob/master/src/Operation/AbstractOperation.php#L21-L24) is probably one of the best "small things" I've seen in PHP recently 😄 . Kudos to you for that!

drupol:
I'm glad you like it and I can't wait to get more feedback!

> Btw [this syntactic sugar](https://github.com/loophp/collection/blob/master/src/Operation/AbstractOperation.php#L21-L24) is probably one of the best "small things" I've seen in PHP recently smile . Kudos to you for that!

This is indeed very helpful :) Thanks for your feedback, comments and contributions!

PR #88, opened 2021-05-20
Update Documentation III

AlexandruGG:
This PR updates more items in the documentation

* [x] Update Usage page
* [x] Update README
* [x] Add missing methods from API docs

Follows #86.


docs/conf.py
```diff
@@ -105,6 +105,8 @@
 # If true, `todo` and `todoList` produce output, else they produce nothing.
 todo_include_todos = False
 
+# Extensions
+extensions = ['sphinx.ext.autosectionlabel']
```
AlexandruGG, line 109: This is based on: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-sections-in-the-same-document

docs/pages/code/e.php
```diff
@@ -21,7 +21,7 @@
 
 $fact = static function (int $number) use ($multiplication): float {
     return Collection::range(1, $number + 1)
-        ->foldleft(
+        ->foldLeft(
```
AlexandruGG, line 24: these are showing as PHPStan errors when running GrumPHP on local, however they don't cause it to fail. Why is that? There are others in the source code too:

```
 ------ ------------------------------------------------------------
  Line   src/Operation/Permutate.php
 ------ ------------------------------------------------------------
  57     Strict comparison using === between array() and array<int,
         T>&nonEmpty will always evaluate to false.
 ------ ------------------------------------------------------------

 ------ ---------------------------------------------------------------------
  Line   src/Operation/Times.php
 ------ ---------------------------------------------------------------------
  52     Anonymous function should return EmptyIterator but return statement
         is missing.
 ------ ---------------------------------------------------------------------

 ------ -------------------------------------------------------------
  Line   src/Operation/Unwindow.php
 ------ -------------------------------------------------------------
  35     Anonymous function should have native return typehint "?T".
  39     Foreach overwrites $value with its value variable.
 ------ -------------------------------------------------------------

**Comment by AlexandruGG** [comment]
There are methods in the [Collection implementation](https://github.com/loophp/collection/blob/master/src/Collection.php) that are not included in the [Collection Interface](https://github.com/loophp/collection/blob/master/src/Contract/Collection.php), and are not mentioned as [methods](https://loophp-collection.readthedocs.io/en/latest/pages/api.html#methods-operations) in the documentation either. A few examples I noticed: `count()`, `current()`. 

Before I added them I wanted to check with you:
1. should they be in the contract? (with their own interface like Countable)
2. should they be in the API docs methods section?

Lastly, a related question: when the library is used in another project, should the _interface_ or the _implementation_ be used in function and method arguments and return types? My instinct says the interface but then the two should be kept in sync otherwise it can result in issues.

File: docs/pages/code/e.php (hunk 2 above)
Comment (drupol) [reply docs/pages/code/e.php:24]: Indeed, the following Grumphp tasks:
- composer_require_checker
- phpstan
- psalm

Are set to not block the checks in case of failure.

This is

File: docs/pages/code/simple.php [hunk 3]
```diff
@@ -71,12 +71,12 @@ static function (string $value, string $key): string {
 
 // Tail
 Collection::fromIterable(range('a', 'z'))
```
drupol, line 73:
```suggestion
Collection::fromIterable(range('a', 'e'))
```

docs/pages/code/simple.php
```diff
@@ -71,12 +71,12 @@ static function (string $value, string $key): string {
...
-    ->tail(3)
+    ->tail()
     ->all(); // [23 => 'x', 24 => 'y', 25 => 'z']
```
drupol, line 75:
```suggestion
    ->all(); // [1 => 'b', 2 => 'c', 3 => 'd', 4 => 'e']
```

docs/pages/code/simple.php
```diff
@@ -71,12 +71,12 @@ static function (string $value, string $key): string {
...
 
 // Reverse
 Collection::fromIterable(range('a', 'z'))
-    ->tail(4)
+    ->tail()
     ->reverse()
     ->all(); // [25 => 'z', 24 => 'y', 23 => 'x', 22 => 'w']
```
drupol, line 81:
```suggestion
    ->all(); // [4 => 'e', 3 => 'd', 2 => 'c', 1 => 'b']
```

docs/pages/code/simple.php
```diff
@@ -71,12 +71,12 @@ static function (string $value, string $key): string {
 
 // Tail
 Collection::fromIterable(range('a', 'z'))
-    ->tail(3)
+    ->tail()
     ->all(); // [23 => 'x', 24 => 'y', 25 => 'z']
 
 // Reverse
 Collection::fromIterable(range('a', 'z'))
```
drupol, line 78:
```suggestion
Collection::fromIterable(range('a', 'e'))
```

drupol (commented):
I added minor changes.

docs/conf.py
drupol: This is great!

drupol:
> @drupol There are methods in the [Collection implementation](https://github.com/loophp/collection/blob/master/src/Collection.php) that are not included in the [Collection Interface](https://github.com/loophp/collection/blob/master/src/Contract/Collection.php), and are not mentioned as [methods](https://loophp-collection.readthedocs.io/en/latest/pages/api.html#methods-operations) in the documentation either. A few examples I noticed: `count()`, `current()`.

Indeed, those are missing and we should fix that.

> Before I added them I wanted to check with you:
> 
>     1. should they be in the contract? (with their own interface like Countable)

`Countable` is already there, but `current` I don't know, I'm still unsure.
I think we should add it for sure, but at the same time, it's not returning a `collection` so, I'm unsure.
What's your opinion on this?

>     2. should they be in the API docs methods section?

Definitely, without hesitation.

> Lastly, a related question: when the library is used in another project, should the _interface_ or the _implementation_ be used in function and method arguments and return types? My instinct says the interface but then the two should be kept in sync otherwise it can result in issues.

I see, this is linked to your previous questions.

I always try to rely on the `interface` all the time and yes, it would make sens to sync everything.

docs/pages/code/e.php
drupol: This is on my TODO list, but feel free to implement it, I'm curious to see how it's done!

AlexandruGG: sure, I've worked with phpstan and psalm baselines before. I'm just not sure how the whole setup works using GrumPHP, but I can look into it

docs/pages/code/duplicate-keys.php
```diff
@@ -57,5 +57,5 @@
  * "(4, b)"
  * "(3, c)"
  * "(4, d)"
- * "(5, e)"
+ * "(5, e)".
```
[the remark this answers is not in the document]
AlexandruGG: Nah it's fine I think it's useful for php docs in general 😄

docs/pages/api.rst
```diff
@@ -1429,6 +1444,46 @@ Signature: ``Collection::permutate(int $size, $value);``
...
+                    yield "{$item}x";
+                }
+            };
+        }
+    };
+
+    Collection::fromIterable(range(1, 5))
+        ->pipe($square, Reverse::of(), $toString, $times())
+        ->all(); // ['25x', '16x', '9x', '4x', '1x']
+
```
AlexandruGG, line 1486: I hope this is a good definition and showcase of `pipe`. Let me know if you think anything else should be included in the description.

Another thing I noticed: it's return type is `self` vs `CollectionInterface` like most other methods: https://github.com/loophp/collection/blob/master/src/Collection.php#L610. Is that just an error or was there a reason for it?

drupol: This is something that I should sort one day. Returning `self` or the `interface`. I haven't found a proper way to do it yet.

Regarding the explanation, it's perfect :)

drupol:
Thanks for your time and dedication, I'm glad and grateful to receive such pull requests!

PR #89, opened 2021-05-20
Code Analysis Fixes - PHPStan

AlexandruGG:
This PR aims to fix current issues reported by PHPStan.

- [x] fixing a few issues without major impact
- [x] adding a baseline file for remaining issues
- [x] making static analysis (Psalm, PHPStan) blocking in GrumPHP

### DIFF
--- a/docs/pages/code/e.php
+++ b/docs/pages/code/e.php
@@ -19,20 +19,17 @@ $addition = static function (float $value1, float $value2): float {
     return $value1 + $value2;
 };
 
-$fact = static function (int $number) use ($multiplication): float {
-    return Collection::range(1, $number + 1)
-        ->foldLeft(
-            $multiplication,
-            1
-        )
+$fact = static function (float $number) use ($multiplication): float {
+    return (float) Collection::range(1, $number + 1)
+        ->foldLeft($multiplication, 1)
         ->current();
 };
 
-$e = static function (int $value) use ($fact): float {
+$e = static function (float $value) use ($fact): float {
     return $value / $fact($value);
 };
 
-$listInt = static function (int $init, callable $succ): Generator {
+$listInt = static function (float $init, callable $succ): Generator {
     yield $init;
 
     while (true) {
@@ -40,7 +37,7 @@ $listInt = static function (int $init, callable $succ): Generator {
     }
 };
 
-$naturals = $listInt(1, static function (int $n): int {
+$naturals = $listInt(1, static function (float $n): float {
     return $n + 1;
 });
 
--- a/spec/loophp/collection/Iterator/RandomIteratorSpec.php
+++ b/spec/loophp/collection/Iterator/RandomIteratorSpec.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace spec\loophp\collection\Iterator;
 
 use ArrayIterator;
+use BadMethodCallException;
 use Exception;
 use loophp\collection\Iterator\RandomIterator;
 use PhpSpec\ObjectBehavior;
@@ -77,6 +78,30 @@ class RandomIteratorSpec extends ObjectBehavior
         if (iterator_to_array($iterator1) === iterator_to_array($this->getWrappedObject())) {
             throw new Exception('Iterator1 is equal to Iterator2');
         }
+    }
+
+    public function it_can_get_current()
+    {
+        $this->beConstructedWith(new ArrayIterator(['a']));
+        $this->valid()->shouldReturn(false);
+        $this->shouldThrow(BadMethodCallException::class)->during('current');
+
+        $this->rewind();
+
+        $this->valid()->shouldReturn(true);
+        $this->current()->shouldReturn('a');
+    }
+
+    public function it_can_get_key()
+    {
+        $this->beConstructedWith(new ArrayIterator(['a' => 1]));
+        $this->valid()->shouldReturn(false);
+        $this->shouldThrow(BadMethodCallException::class)->during('key');
+
+        $this->rewind();
+
+        $this->valid()->shouldReturn(true);
+        $this->key()->shouldReturn('a');
     }
 
     public function it_can_get_the_innerIterator()
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -376,12 +376,12 @@ final class Collection implements CollectionInterface
      *
      * @param mixed ...$parameters
      */
-    public static function fromCallable(callable $callable, ...$parameters): self
+    public static function fromCallable(callable $callable, ...$parameters): CollectionInterface
     {
         return new self($callable, ...$parameters);
     }
 
-    public static function fromFile(string $filepath): self
+    public static function fromFile(string $filepath): CollectionInterface
     {
         return new self(
             /**
@@ -400,7 +400,7 @@ final class Collection implements CollectionInterface
      * @param iterable<mixed> $iterable
      * @psalm-param iterable<NewTKey, NewT> $iterable
      */
-    public static function fromIterable(iterable $iterable): self
+    public static function fromIterable(iterable $iterable): CollectionInterface
     {
         return new self(
             /**
@@ -416,7 +416,7 @@ final class Collection implements CollectionInterface
     /**
      * @param resource $resource
      */
-    public static function fromResource($resource): self
+    public static function fromResource($resource): CollectionInterface
     {
         return new self(
             /**
@@ -430,7 +430,7 @@ final class Collection implements CollectionInterface
         );
     }
 
-    public static function fromString(string $string, string $delimiter = ''): self
+    public static function fromString(string $string, string $delimiter = ''): CollectionInterface
     {
         return new self(
             /**
@@ -607,7 +607,7 @@ final class Collection implements CollectionInterface
         return new self(Permutate::of(), $this->getIterator());
     }
 
-    public function pipe(callable ...$callables): self
+    public function pipe(callable ...$callables): CollectionInterface
     {
         return new self(Pipe::of()(...$callables), $this->getIterator());
     }
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -376,12 +376,12 @@ final class Collection implements CollectionInterface
      *
      * @param mixed ...$parameters
      */
-    public static function fromCallable(callable $callable, ...$parameters): CollectionInterface
+    public static function fromCallable(callable $callable, ...$parameters): self
     {
         return new self($callable, ...$parameters);
     }
 
-    public static function fromFile(string $filepath): CollectionInterface
+    public static function fromFile(string $filepath): self
     {
         return new self(
             /**
@@ -400,7 +400,7 @@ final class Collection implements CollectionInterface
      * @param iterable<mixed> $iterable
      * @psalm-param iterable<NewTKey, NewT> $iterable
      */
-    public static function fromIterable(iterable $iterable): CollectionInterface
+    public static function fromIterable(iterable $iterable): self
     {
         return new self(
             /**
@@ -416,7 +416,7 @@ final class Collection implements CollectionInterface
     /**
      * @param resource $resource
      */
-    public static function fromResource($resource): CollectionInterface
+    public static function fromResource($resource): self
     {
         return new self(
             /**
@@ -430,7 +430,7 @@ final class Collection implements CollectionInterface
         );
     }
 
-    public static function fromString(string $string, string $delimiter = ''): CollectionInterface
+    public static function fromString(string $string, string $delimiter = ''): self
     {
         return new self(
             /**
--- a/src/Contract/Operation/Unfoldable.php
+++ b/src/Contract/Operation/Unfoldable.php
@@ -18,7 +18,7 @@ interface Unfoldable
      * @psalm-template TKey of array-key
      * @psalm-template T
      *
-     * @psalm-param callable(T...): array<TKey, T> $callback
+     * @psalm-param callable(mixed|T...): (mixed|array<TKey, T>) $callback
      * @psalm-param T ...$parameters
      *
      * @psalm-return \loophp\collection\Collection<TKey, T>
--- a/src/Iterator/RandomIterator.php
+++ b/src/Iterator/RandomIterator.php
@@ -9,8 +9,10 @@ declare(strict_types=1);
 
 namespace loophp\collection\Iterator;
 
+use BadMethodCallException;
 use Iterator;
 
+use function assert;
 use const PHP_INT_MAX;
 use const PHP_INT_MIN;
 
@@ -51,6 +53,12 @@ final class RandomIterator extends ProxyIterator
 
     public function current()
     {
+        if (!$this->valid()) {
+            throw new BadMethodCallException("Cannot call 'current()' on invalid RandomIterator.");
+        }
+
+        assert(null !== $this->key);
+
         $keyValueTuple = $this->getNextItemAtKey($this->key);
 
         return $keyValueTuple[1];
@@ -58,6 +66,12 @@ final class RandomIterator extends ProxyIterator
 
     public function key()
     {
+        if (!$this->valid()) {
+            throw new BadMethodCallException("Cannot call 'key()' on invalid RandomIterator.");
+        }
+
+        assert(null !== $this->key);
+
         $keyValueTuple = $this->getNextItemAtKey($this->key);
 
         return $keyValueTuple[0];
@@ -82,7 +96,7 @@ final class RandomIterator extends ProxyIterator
 
     public function valid(): bool
     {
-        return [] !== $this->indexes;
+        return [] !== $this->indexes && null !== $this->key;
     }
 
     /**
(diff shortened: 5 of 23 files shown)


docs/pages/code/e.php
```diff
@@ -19,28 +19,25 @@
...
-$fact = static function (int $number) use ($multiplication): float {
-    return Collection::range(1, $number + 1)
-        ->foldLeft(
-            $multiplication,
-            1
-        )
+$fact = static function (float $number) use ($multiplication): float {
+    return (float) Collection::range(1, $number + 1)
+        ->foldLeft($multiplication, 1)
         ->current();
```
AlexandruGG, line 25: the `(float)` cast here should not be needed but it is because of the [return type on the interface](https://github.com/loophp/collection/blob/master/src/Contract/Operation/FoldLeftable.php#L31):
`@psalm-return \loophp\collection\Collection<TKey, T|null>`

I feel this should be tied to the return type of the callable passed to the method. Only if the callable returns `null` will the method return type be `null`, otherwise it will be `T`. However, I'm not too sure how to link the two together

src/Collection.php
```diff
@@ -376,12 +376,12 @@ public function frequency(): CollectionInterface
      *
      * @param mixed ...$parameters
      */
-    public static function fromCallable(callable $callable, ...$parameters): self
+    public static function fromCallable(callable $callable, ...$parameters): CollectionInterface
```
AlexandruGG, line 379: will there be any issues with this change? I'm thinking the only potential issues might be due to the interface and the class not being fully in sync when it comes to annotations, but functionality should be impacted

src/Contract/Operation/Unfoldable.php
```diff
@@ -18,7 +18,7 @@ interface Unfoldable
      * @psalm-template TKey of array-key
      * @psalm-template T
      *
-     * @psalm-param callable(T...): array<TKey, T> $callback
+     * @psalm-param callable(mixed|T...): (mixed|array<TKey, T>) $callback
```
AlexandruGG, line 21: this is based on the Operation annotation: https://github.com/loophp/collection/blob/master/src/Operation/Unfold.php#L38

AlexandruGG:
@drupol I noticed the Psalm level set up in the config is 7, which is very lenient: https://psalm.dev/docs/running_psalm/error_levels/. A bit confusingly, the levels go the other way in PHPStan's case, where higher is stricter 😄 .

Was Psalm added later to project? Or how come it's level is not very strict compared to PHPStan?

spec/loophp/collection/Iterator/RandomIteratorSpec.php
```diff
@@ -79,6 +80,30 @@ public function it_can_build_an_iterator_without_a_random_seed()
         }
     }
 
+    public function it_can_get_current()
+    {
+        $this->beConstructedWith(new ArrayIterator(['a']));
+        $this->valid()->shouldReturn(false);
+        $this->shouldThrow(BadMethodCallException::class)->during('current');
```
AlexandruGG, line 87: doing this before would've resulted in a `TypeError`

src/Iterator/RandomIterator.php
```diff
@@ -51,13 +53,25 @@ public function __construct(Iterator $iterator, ?int $seed = null)
...
         return $keyValueTuple[1];
     }
 
     public function key()
     {
+        if (!$this->valid()) {
+            throw new BadMethodCallException("Cannot call 'key()' on invalid RandomIterator.");
+        }
+
+        assert(null !== $this->key);
```
AlexandruGG, line 73: have you considered adding an assertion library to this project, like https://github.com/webmozarts/assert? Together with the [phpstan extension](https://github.com/phpstan/phpstan-webmozart-assert) it can be a powerful tool

AlexandruGG: the assertion is needed here to show code analysers that `$this->key` cannot be null after this point (we've already checked that in `valid()`)

docs/pages/code/e.php
drupol: Yes, I'm sharing the same feeling. This has always been an issue: "How to properly type those reduction operations".
To be honest, I do not pretend to have yet the proper solution.

src/Collection.php
drupol: I always use `self` in conjunction of a static constructor. To me, it makes more sense.

Could you please give me more details on this change? What's the reason?

src/Contract/Operation/Unfoldable.php
drupol: Yep definitely something out of sync that needs to be fixed.

spec/loophp/collection/Iterator/RandomIteratorSpec.php
drupol: I will test during the day, curious to see and understand those changes.

However, maybe we should create a PR only with those changes because they are not related to static analysis but merely to iterators.

src/Iterator/RandomIterator.php
drupol: No I never considered it.

There is a simple reason to that. I want this library to have no dependency. I want it very simple and out of any external deps.

src/Collection.php
AlexandruGG: I think the main reason was due to current differences in the typing of the interface vs the class methods, it can cause issues with code analysis. But what you're saying makes sense actually so I think I'll revert this and those differences will just need to be fixed

docs/pages/code/e.php
AlexandruGG: No worries, I'll look into it some more. We might need a [type-specifying extension](https://phpstan.org/developing-extensions/type-specifying-extensions) for some of these methods. PHPStan is not running with a config file at the moment, right?

spec/loophp/collection/Iterator/RandomIteratorSpec.php
AlexandruGG: It's related to both the iterator behaviour and fixing static analysis. Effectively the below error is warning us of a TypeError, which happens if `key` or `current` is called when the key is null (out of bounds).

``` ------ --------------------------------------------------------------------------------------------
  Line   RandomIterator.php
 ------ --------------------------------------------------------------------------------------------
  54     Parameter #1 $key of method loophp\collection\Iterator\RandomIterator<TKey,T of
         string>::getNextItemAtKey() expects int, int|null given.
  61     Parameter #1 $key of method loophp\collection\Iterator\RandomIterator<TKey,T of
         string>::getNextItemAtKey() expects int, int|null given.
```

Example which causes a type error:

```PHP
$it = new ArrayIterator([1]);
$rand = new RandomIterator($it);
$rand->rewind();
$rand->next();

var_dump($rand->key());
```

```
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to loophp\collection\Iterator\RandomIterator::getNextItemAtKey() must be of the type int, null given, called in /Users/alex/dev/collections/forked/collection/src/Iterator/RandomIterator.php on line 76 and defined in /Users/alex/dev/collections/forked/collection/src/Iterator/RandomIterator.php:109
```

Another example which causes an infinite loop:

```PHP
$it = new ArrayIterator([1]);
$rand = new RandomIterator($it);

var_dump($rand->current());

```

```
Notice: Undefined offset: 0 in /Users/alex/dev/collections/forked/collection/src/Iterator/RandomIterator.php on line 115
PHP Notice:  Undefined offset: 0 in /Users/alex/dev/collections/forked/collection/src/Iterator/RandomIterator.php on line 115```

Comment (AlexandruGG) [reply spec/loophp/collection/Iterator/RandomIteratorSpec.php:87]: The `BadMethodCallException` helps protect against those and signals to the caller that they're doing something wrong.

File: src/Iterator/RandomIterator.php (hunk 5 above)
Comment (AlexandruGG) [reply src/Iterator/RandomIterator.php:74]: Yep I thought that might've been the reason 😄 . Makes sense, there are ways to get around it 👍

File: docs/pages/code/e.php (hunk 1 above)
Comment (drupol) [reply docs/pages/code/e.php:25]: No, nothing.

Run `./vendor/bin/grumphp run --tasks=phpstan,psalm -vvv` to see the commands that are used.

File: src/Collection.php (hunk 2 above)
Comment (drupol) [reply src/Collection.php:379]: Fair enough.

File: spec/loophp/collection/Iterator/RandomIteratorSpec.php (hunk 4 above)
Comment (drupol) [reply spec/loophp/collection/Iterator/RandomIteratorSpec.php:87]: Allrighty! Good.

File: docs/pages/code/e.php (hunk 1 above)
Comment (AlexandruGG) [reply docs/pages/code/e.php:25]: I found this, potentially we'd want the `foldLeft` and `reduction` typed more similarly to the `reduce` here? https://arnaud.le-blanc.net/post/phpstan-generics.html

Basically saying the return type will be the type of the initial value, which should correspond to the type returned by the callback

Comment (drupol) [reply docs/pages/code/e.php:25]: I see.

I started to work on a subpackage here, I guess you'll be interested to see the PSALM/PHPStan annotations: https://github.com/loophp/fpt

Comment (AlexandruGG) [reply docs/pages/code/e.php:25]: Ah yes, this is the correct way: https://github.com/loophp/fpt/blob/master/src/Reduce.php#L32.

I also reached that conclusion 😄 : https://phpstan.org/r/f57b8607-c972-4d1d-bdf3-3facd22f076e

Comment (AlexandruGG) [reply docs/pages/code/e.php:25]: I am reluctant to embark on significant changes of the annotations in this project until we have some "test files" for the code analysis.

What do you think of that? Basically having a dedicated folder where we just try to use the library like a user would, and ensure phpstan+psalm pass at close to the maximum levels. I'm aware we have the "examples"  docs but those are more just for showcasing the library whereas this would be part of the CI suite and errors should not be allowed

Comment (drupol) [reply docs/pages/code/e.php:25]: Same tests with PSalm: https://psalm.dev/r/f07db5da95

The more tests we have, the better, so I'm all :+1: for this!

**Comment by AlexandruGG** [comment]
@drupol I decided to err on the side of caution with this and so only fixed some of the things what were not very problematic. I tried diving into the issues in `Collection.php`, however fixing the types there leads to a lot more issues because PHPStan is then able to better infer the types in other places 😄 .

I added a [baseline file](https://github.com/loophp/collection/pull/89/files#diff-995edee38ad4f8387e58ebd52c31bcc04c56cc2448d331b1cf5e0b35c57b9efaR1-R2) so now these are clearly visible and can be tackled separately.

How do we make PHPStan+Psalm blocking in GrumPHP? Do we do it via a parameter in this project which overwrites the [drup-conventions parameter](https://github.com/drupol/php-conventions/blob/master/config/php73/grumphp.yml#L92)? Or do we change the one in that repository? (not sure how it affects other projects of yours)

**Comment by AlexandruGG** [comment]
> @drupol I noticed the Psalm level set up in the config is 7, which is very lenient: https://psalm.dev/docs/running_psalm/error_levels/. A bit confusingly, the levels go the other way in PHPStan's case, where higher is stricter 😄 .
> 
> Was Psalm added later to project? Or how come it's level is not very strict compared to PHPStan?


Trying the library in a project that uses a stricter Psalm level results in errors 😢 

`
ERROR: InvalidArgument - tests/Util/Command/CheckCrapScore.php:120:57 - Argument 1 of loophp\collection\Collection::map expects callable(int, array-key, Iterator<array-key, mixed>):V:loophp\collection\Contract\Operation\Mapable as mixed, pure-Closure(int):int provided (see https://psalm.dev/004)`
        
`$collection = Collection::fromIterable([1, 2, 3])->map(static fn(int $value): int => $value * 2);`

I scratched my head with this one a bit but still couldn't see what the issue was 😅

**Comment by drupol** [comment]
> I added a [baseline file](https://github.com/loophp/collection/pull/89/files#diff-995edee38ad4f8387e58ebd52c31bcc04c56cc2448d331b1cf5e0b35c57b9efaR1-R2) so now these are clearly visible and can be tackled separately.

Excellent, I really like that!

> How do we make PHPStan+Psalm blocking in GrumPHP? Do we do it via a parameter in this project which overwrites the [drup-conventions parameter](https://github.com/drupol/php-conventions/blob/master/config/php73/grumphp.yml#L92)? Or do we change the one in that repository? (not sure how it affects other projects of yours)

Like this:

```diff
diff --git a/grumphp.yml b/grumphp.yml
index a5dba304..0bb8a40e 100644
--- a/grumphp.yml
+++ b/grumphp.yml
@@ -5,6 +5,7 @@ parameters:
   tasks.phpcsfixer.config: .php-cs-fixer.dist.php
   tasks.license.holder: Pol Dellaiera
   tasks.license.date_from: 2019
+  tasks.phpstan.blocking: true
   extra_tasks:
     phpspec:
       verbose: true
```

drupol:
> Trying the library in a project that uses a stricter Psalm level results in errors cry

Yeah I'm aware of this :(
 
> ` ERROR: InvalidArgument - tests/Util/Command/CheckCrapScore.php:120:57 - Argument 1 of loophp\collection\Collection::map expects callable(int, array-key, Iterator<array-key, mixed>):V:loophp\collection\Contract\Operation\Mapable as mixed, pure-Closure(int):int provided (see https://psalm.dev/004)`
> 
> `$collection = Collection::fromIterable([1, 2, 3])->map(static fn(int $value): int => $value * 2);`
> 
> I scratched my head with this one a bit but still couldn't see what the issue was sweat_smile

The `callable` takes 3 arguments.

- The value
- The key
- The iterator.

Try with something like this:

`static fn(int $value, $key, Iterator $iterator): int => $value * 2`

AlexandruGG:
> > Trying the library in a project that uses a stricter Psalm level results in errors cry
> 
> Yeah I'm aware of this :(
> 
> > ` ERROR: InvalidArgument - tests/Util/Command/CheckCrapScore.php:120:57 - Argument 1 of loophp\collection\Collection::map expects callable(int, array-key, Iterator<array-key, mixed>):V:loophp\collection\Contract\Operation\Mapable as mixed, pure-Closure(int):int provided (see https://psalm.dev/004)`
> > `$collection = Collection::fromIterable([1, 2, 3])->map(static fn(int $value): int => $value * 2);`
> > I scratched my head with this one a bit but still couldn't see what the issue was sweat_smile
> 
> The `callable` takes 3 arguments.
> 
> * The value
> * The key
> * The iterator.
> 
> Try with something like this:
> 
> `static fn(int $value, $key, Iterator $iterator): int => $value * 2`

Unfortunately I don't think that's the issue as the map should be able to be used without those too. Adding them still errors:

`ERROR: InvalidArgument - tests/Util/Command/CheckCrapScore.php:121:57 - Argument 1 of loophp\collection\Collection::map expects callable(int, int, Iterator<TKey:loophp\collection\Collection as array-key, T:loophp\collection\Collection as mixed>):V:loophp\collection\Contract\Operation\Mapable as mixed, pure-Closure(int, int, Iterator<TKey:loophp\collection\Collection as array-key, T:loophp\collection\Collection as mixed>):int provided (see https://psalm.dev/004)`

`$collection = Collection::fromIterable([1, 2, 3])->map(static fn(int $value, int $key, Iterator $iterator): int => $value * 2);`

I think it has to do with the `map` return type instead, that part in the error message looks like it's different.

drupol:
This is something that we need to sort out. I know that annotations are not well done, but this is a huge work. I haven't been able to find the right balance.

AlexandruGG:
> This is something that we need to sort out. I know that annotations are not well done, but this is a huge work. I haven't been able to find the right balance.

I know, it'll be hard to fix them with the code already in place because there are interdependencies. 
I want to try the suggestion I made [here](https://github.com/loophp/collection/pull/89#discussion_r636743099) in a future PR and adjust the Psalm level so those errors become visible.

What would be the best place to put that code? I don't think it quite fits in docs, src, spec, or tests (also the last 2 are excluded from analysis). Should we create a new folder `usage`?

AlexandruGG:
> > This is something that we need to sort out. I know that annotations are not well done, but this is a huge work. I haven't been able to find the right balance.
> 
> I know, it'll be hard to fix them with the code already in place because there are interdependencies.
> I want to try the suggestion I made [here](https://github.com/loophp/collection/pull/89#discussion_r636743099) in a future PR and adjust the Psalm level so those errors become visible.
> 
> What would be the best place to put that code? I don't think it quite fits in docs, src, spec, or tests (also the last 2 are excluded from analysis). Should we create a new folder `usage`?

Update on this, I think a folder called `static-analysis` would be more descriptive. A colleague alerted me to the fact that this pattern is commonly used:
- https://github.com/moneyphp/money/tree/master/static-analysis
- https://github.com/Ocramius/ProxyManager/tree/2.12.x/tests/static-analysis

drupol:
Yeah I did that for this: https://github.com/loophp/combinator

AlexandruGG:
> Yeah I did that for this: https://github.com/loophp/combinator

Ah nice! Yeah this library needs that too. Don't keep the best stuff for your other projects 😜 😄

drupol:
I have a local branch of loophp/collection completely rewritten with loophp/fpt :)

But this is for later, much later!

AlexandruGG:
Let me know if there's anything else to address on this, it's ready from my side :)

drupol:
You're on a roll ! Thanks again !!!

PR #90, opened 2021-05-21
Introduce Static Analysis Checking System

AlexandruGG:
This PR aims to kickstart improvements to static analysis
Follows #89.

### Overall
* [x] Increase Psalm strictness & add baseline
* [x] Introduce `tests/static-analysis`
* [x] Fixing some issues where applicable

### Checks added
* [x] fromFile 
* [x] fromResource
* [x] fromString

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,14 +1,13 @@
 <?xml version="1.0"?>
 <psalm
-    errorLevel="7"
-    totallyTyped="true"
-    resolveFromConfigFile="true"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
+    errorBaseline="psalm-baseline.xml"
 >
     <projectFiles>
         <directory name="src" />
+        <directory name="tests" />
         <ignoreFiles>
             <directory name="vendor" />
             <directory name="spec" />
--- a/spec/loophp/collection/CollectionSpec.php
+++ b/spec/loophp/collection/CollectionSpec.php
@@ -21,6 +21,8 @@ use loophp\collection\Collection;
 use loophp\collection\Contract\Operation;
 use loophp\collection\Operation\AbstractOperation;
 use OutOfBoundsException;
+use PhpSpec\Exception\Example\FailureException;
+use PhpSpec\Exception\Example\MatcherException;
 use PhpSpec\ObjectBehavior;
 use stdClass;
 use const INF;
@@ -86,7 +88,7 @@ class CollectionSpec extends ObjectBehavior
         ];
 
         if ($stack !== $expected) {
-            throw new Exception('The expected value does not match.');
+            throw new MatcherException('The expected value does not match.');
         }
 
         $stack = [];
@@ -107,7 +109,7 @@ class CollectionSpec extends ObjectBehavior
         ];
 
         if ($stack !== $expected) {
-            throw new Exception('The expected value does not match.');
+            throw new MatcherException('The expected value does not match.');
         }
 
         $stack = [];
@@ -138,7 +140,7 @@ class CollectionSpec extends ObjectBehavior
         ];
 
         if ($stack !== $expected) {
-            throw new Exception('The expected value does not match.');
+            throw new MatcherException('The expected value does not match.');
         }
 
         $stack = [];
@@ -175,7 +177,7 @@ class CollectionSpec extends ObjectBehavior
         ];
 
         if ($stack !== $expected) {
-            throw new Exception('The expected value does not match.');
+            throw new MatcherException('The expected value does not match.');
         }
     }
 
@@ -838,7 +840,7 @@ class CollectionSpec extends ObjectBehavior
             ->shouldIterateAs($input);
 
         if (5 !== $count) {
-            throw new Exception('Invalid count1');
+            throw new FailureException('Invalid count1');
         }
 
         $count = 0;
@@ -852,7 +854,7 @@ class CollectionSpec extends ObjectBehavior
             ->shouldIterateAs($input);
 
         if (0 !== $count) {
-            throw new Exception('Invalid count2');
+            throw new FailureException('Invalid count2');
         }
 
         $callback = static function (string $name, $key, $value) use (&$count) {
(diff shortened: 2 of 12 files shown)


psalm.xml
```diff
@@ -1,14 +1,13 @@
 <?xml version="1.0"?>
 <psalm
-    errorLevel="7"
-    totallyTyped="true"
-    resolveFromConfigFile="true"
```
AlexandruGG, line 5: going with the default strictness level of 2. The stricter level 1 is too hardcore with too many `Mixed*` error messages: https://psalm.dev/docs/running_psalm/error_levels/

spec/loophp/collection/Iterator/ResourceIteratorSpec.php
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public function it_is_initializable(): void
+    {
+        $file = fopen(__DIR__ . '/../../../fixtures/sample.txt', 'rb');
+
+        $this->beConstructedWith($file);
+
+        $this->shouldHaveType(ResourceIterator::class);
+    }
```
AlexandruGG, line 46: I'm not too familiar with PHPSpec, though I use PHPUnit + Prophecy so some things look familiar 😄 . Let me know if this looks good or I could be doing something better based on best practices.

I wanted to ensure that the change I made [here](https://github.com/loophp/collection/pull/90/files#diff-be52b0e1d1956fe3eefb225a24ccb14b0ffd5c769563b74084f74f892d80678fR33) is tested

tests/static-analysis/fromFile.php
```diff
@@ -0,0 +1,21 @@
...
+use loophp\collection\Collection;
+
+/**
+ * @param Collection<int, string> $collection
+ */
+function check(Collection $collection): void
+{
+}
+
+check(Collection::fromFile('https://loripsum.net/api')->limit(25));
```
AlexandruGG, line 21: let me know if this makes sense, all I wanted to ensure is that the return type of `fromFile` is properly registered and checked by Psalm & PHPStan as `Collection<int, string>`

psalm.xml
drupol: Sounds good for me!

spec/loophp/collection/Iterator/ResourceIteratorSpec.php
drupol: Seems to be nice at first sight!

drupol:
It seems that `::fromFile()` is a new constructor.

I've always been reluctant to add it because I haven't found a proper way to do the `fclose()` at the end. Do you have a clue??

I'm not so keen on leaving the resource open without closing them.

AlexandruGG:
> It seems that `::fromFile()` is a new constructor.
> 
> I've always been reluctant to add it because I haven't found a proper way to do the `fclose()` at the end. Do you have a clue??
> 
> I'm not so keen on leaving the resource open without closing them.

Challenge accepted! 😄 

Let me know if this looks good: https://github.com/loophp/collection/pull/90/commits/e000805633c45eaff0a214bc84121148100d2733

spec/loophp/collection/CollectionSpec.php
```diff
@@ -86,7 +88,7 @@ static function ($item) use (&$stack): bool {
         ];
 
         if ($stack !== $expected) {
-            throw new Exception('The expected value does not match.');
+            throw new MatcherException('The expected value does not match.');
```
AlexandruGG, line 91: using the PHPSpec matchers will result in specs being reported as `failed` vs `broken` if the exception is thrown

AlexandruGG:
> > It seems that `::fromFile()` is a new constructor.
> > I've always been reluctant to add it because I haven't found a proper way to do the `fclose()` at the end. Do you have a clue??
> > I'm not so keen on leaving the resource open without closing them.
> 
> Challenge accepted! 😄
> 
> Let me know if this looks good: [e000805](https://github.com/loophp/collection/commit/e000805633c45eaff0a214bc84121148100d2733)

I realised that this iterator is also used in [fromResource](https://github.com/loophp/collection/blob/master/src/Collection.php#L428), where the user already provides a resource. We might not want to close it in that case as it should be up to the caller, so I'm thinking to add a boolean parameter to ResourceIterator with default value `false` that should be used to check if the resource should be closed. Does that sound good?

tests/static-analysis/fromFile.php
drupol: This seems to be the proper way indeed.

spec/loophp/collection/CollectionSpec.php
drupol: Nice !

drupol (dismissed):
> I realised that this iterator is also used in [fromResource](https://github.com/loophp/collection/blob/master/src/Collection.php#L428), where the user already provides a resource. We might not want to close it in that case as it should be up to the caller, so I'm thinking to add a boolean parameter to ResourceIterator with default value `false` that should be used to check if the resource should be closed. Does that sound good?

This is exactly why it is done like that, I let the user close its `resource`.

In overall, I'm ok with the pull request, feel free to submit it!

AlexandruGG:
> > I realised that this iterator is also used in [fromResource](https://github.com/loophp/collection/blob/master/src/Collection.php#L428), where the user already provides a resource. We might not want to close it in that case as it should be up to the caller, so I'm thinking to add a boolean parameter to ResourceIterator with default value `false` that should be used to check if the resource should be closed. Does that sound good?
> 
> This is exactly why it is done like that, I let the user close it's `resource`.
> 
> In overall, I'm ok with the pull request, feel free to submit it!

Ok so we don't want `ResourceIterator` to always close the resource. I will add that parameter to ensure that we only close it in `fromFile` and not in `fromResource`.

Also, I'll add the static analysis check for `fromResource` and after that it'll be ready for review :)

drupol:
Maybe a new extra parameter to ResourceIterator ?

src/Collection.php
```diff
@@ -415,15 +415,14 @@ public static function fromIterable(iterable $iterable): self
 
     /**
      * @param resource $resource
+     *
+     * @return self<int, string>
```
AlexandruGG, line 419: @drupol I see a mix of annotations in this project, but I'm not sure why that is. Should we use the psalm ones or the "normal" ones everywhere?

A. `@param`, `@return`, `@template`
B.  `@psalm-param`, `@psalm-return`, `@psalm-template`

For the analysers it doesn't matter because they support both. And in an IDE like PHPStorm / IntelliJ both are properly highlighted.

I personally favour **A** just because of a couple small reasons:
- less characters to write. And easier to mentally parse when reading 😄 
- doesn't tie the code to a specific developer tool (though, again, doesn't really matter because both are supported)

drupol: This is also on my Todo list! Option A for the win :)

drupol (approved):
Excellent!

PR #94, opened 2021-05-23
SA Checks: fromCallable

AlexandruGG:
Continuing the static analysis checks, this PR adds `fromCallable`.
- [x] adding check
- [x] improving tests
- [x] improving type hints

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -6,17 +6,7 @@ parameters:
 			path: docs/pages/code/operations/since.php
 
 		-
-			message: "#^Method loophp\\\\collection\\\\Collection\\:\\:all\\(\\) should return array\\<TKey, T\\> but returns array\\.$#"
-			count: 1
-			path: src/Collection.php
-
-		-
 			message: "#^Method loophp\\\\collection\\\\Collection\\:\\:empty\\(\\) return type with generic interface loophp\\\\collection\\\\Contract\\\\Collection does not specify its types\\: TKey, T$#"
-			count: 1
-			path: src/Collection.php
-
-		-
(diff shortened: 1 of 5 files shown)


phpstan-baseline.neon
```diff
@@ -5,33 +5,18 @@ parameters:
...
 			path: src/Collection.php
 
 		-
-			message: "#^Method loophp\\\\collection\\\\Collection\\:\\:getIterator\\(\\) return type has no value type specified in iterable type Iterator\\.$#"
-			count: 1
-			path: src/Collection.php
-
-		-
-			message: "#^Return type \\(Iterator\\) of method loophp\\\\collection\\\\Collection\\:\\:getIterator\\(\\) should be covariant with return type \\(Iterator\\<mixed, mixed\\>\\) of method loophp\\\\collection\\\\Contract\\\\Collection\\<TKey,T\\>\\:\\:getIterator\\(\\)$#"
+			message: "#^Parameter \\#1 \\$callable of class loophp\\\\collection\\\\Collection constructor expects callable\\(\\.\\.\\.mixed\\)\\: iterable\\<int, string\\>, Closure\\(string, string\\)\\: loophp\\\\collection\\\\Iterator\\\\StringIterator\\<mixed, string\\> given\\.$#"
```
AlexandruGG, line 19: This seems to be an issue with PHPStan, I've reported it here: https://github.com/phpstan/phpstan/issues/5060

drupol (approved):
Super great ! Thanks !

AlexandruGG:
@drupol I just realised that in CI PHPStan doesn't run on the `tests/static-analysis` directory because it's ignored here: https://github.com/drupol/php-conventions/blob/master/config/php73/grumphp.yml#L85. I've been using it directly rather than via GrumPHP on my local so far and overlooked that.

How can we override that config for GrumPHP? Do we need to have a `tasks.phpstan.ignore_patterns` entry in the project's config where we list all paths again apart from the tests one?

drupol:
Yes that's it.

PR #95, opened 2021-05-24
Update annotations - remove @psalm prefix

drupol:
This PR

* [x] Replace @psalm-template with @template.
* [x] Replace @psalm-param with @param.
* [x] Replace @psalm-var with @var.
* [x] Replace @psalm-return with @return.
* [x] Fix code-style.

PR #96, opened 2021-05-24
SA Checks: fromIterable, empty

AlexandruGG:
This PR adds to the static analysis checks: `fromIterable`, `empty`.

* [x] add checks
* [x] add PHPStan analysis for `tests/` in GrumPHP
* [x] a couple of tweaks to `IterableIterator`
* [x] fixing a bunch of static analysis issues around `Unpack` -> surfaced by the `squash` operation now that `fromIterable` has the right type hints
* [x] added some more PHPStan baselines for specific purposes

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,2 +1,4 @@
 includes:
 	- phpstan-baseline.neon
+	- phpstan-docs-baseline.neon
+	- phpstan-unsupported-baseline.neon
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -291,13 +291,17 @@ final class Collection implements CollectionInterface
     /**
      * Create a new instance with no items.
      *
-     * @psalm-template NewTKey
-     * @psalm-template NewTKey of array-key
-     * @psalm-template NewT
+     * @template NewTKey of array-key
+     * @template NewT
+     *
+     * @return self<NewTKey, NewT>
      */
     public static function empty(): CollectionInterface
     {
-        return self::fromIterable([]);
+        /** @var array<NewTKey, NewT> $emptyArray */
+        $emptyArray = [];
+
+        return self::fromIterable($emptyArray);
     }
 
     public function every(callable ...$callbacks): CollectionInterface
@@ -366,8 +370,8 @@ final class Collection implements CollectionInterface
     }
 
     /**
-     * @psalm-template NewTKey of array-key
-     * @psalm-template NewT
+     * @template NewTKey of array-key
+     * @template NewT
      *
      * @param callable(mixed ...$parameters): iterable<NewTKey, NewT> $callable
      * @param mixed ...$parameters
@@ -391,21 +395,16 @@ final class Collection implements CollectionInterface
     }
 
     /**
-     * @psalm-template NewTKey
-     * @psalm-template NewTKey of array-key
-     * @psalm-template NewT
+     * @template NewTKey of array-key
+     * @template NewT
      *
-     * @param iterable<mixed> $iterable
-     * @psalm-param iterable<NewTKey, NewT> $iterable
+     * @param iterable<NewTKey, NewT> $iterable
+     *
+     * @return self<NewTKey, NewT>
      */
     public static function fromIterable(iterable $iterable): self
     {
         return new self(
-            /**
-             * @psalm-param iterable<TKey, T> $iterable
-             *
-             * @psalm-return Iterator<TKey, T>
-             */
             static fn (iterable $iterable): Iterator => new IterableIterator($iterable),
             $iterable
         );
--- a/src/Contract/Operation/Unpackable.php
+++ b/src/Contract/Operation/Unpackable.php
@@ -12,14 +12,16 @@ namespace loophp\collection\Contract\Operation;
 use loophp\collection\Contract\Collection;
 
 /**
- * @psalm-template TKey
- * @psalm-template TKey of array-key
- * @psalm-template T
+ * @template TKey of int
+ * @template T of array{0: NewTKey, 1: NewT}
+ *
+ * @template NewTKey of array-key
+ * @template NewT
  */
 interface Unpackable
 {
     /**
-     * @psalm-return \loophp\collection\Collection<TKey, T>
+     * @return \loophp\collection\Collection<NewTKey, NewT>
      */
     public function unpack(): Collection;
 }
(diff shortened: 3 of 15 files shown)


src/Collection.php
```diff
@@ -291,13 +291,17 @@ public function duplicate(): CollectionInterface
...
+     *
+     * @return self<NewTKey, NewT>
      */
     public static function empty(): CollectionInterface
     {
-        return self::fromIterable([]);
+        /** @var array<NewTKey, NewT> $emptyArray */
+        $emptyArray = [];
+
+        return self::fromIterable($emptyArray);
```
AlexandruGG, line 304: this is the only way to get around having template parameters for the empty array, otherwise PHPStan complains because it cannot figure out the template types:
```
  301    Method loophp\collection\Collection::empty() should
         return loophp\collection\Collection<NewTKey of
         (int|string), NewT> but returns
         loophp\collection\Collection<*NEVER*, *NEVER*>.

File: src/Contract/Operation/Unpackable.php [hunk 2]
```diff
@@ -12,14 +12,16 @@
...
 
 /**
- * @psalm-template TKey
- * @psalm-template TKey of array-key
- * @psalm-template T
+ * @template TKey of int
+ * @template T of array{0: NewTKey, 1: NewT}
+ *
+ * @template NewTKey of array-key
+ * @template NewT
```
AlexandruGG, line 19: I literally must've spent over an hour getting these type hints right 😄

AlexandruGG: but it's beautiful when the static analysis finally turns green ✅ 😆

src/Iterator/IterableIterator.php
```diff
@@ -22,20 +22,12 @@
...
         $this->iterator = new ClosureIterator(
-            /**
-             * @psalm-param iterable<TKey, T> $iterable
-             */
-            static function (iterable $iterable): Generator {
-                foreach ($iterable as $key => $value) {
-                    yield $key => $value;
-                }
-            },
+            static fn (iterable $iterable): Generator => yield from $iterable,
```
AlexandruGG, line 30: This is equivalent just shorter, right? From the tests I did it seemed to be

src/Contract/Operation/Unpackable.php
AlexandruGG: damn looks like this is not actually supported yet in PHPStan 😄. Not sure why it was showing all good in another file

AlexandruGG: trying to work around it makes this worse and incorrect, so I think I'll keep this. Psalm supports it and it passes, and PHPStan should support it soon: https://github.com/phpstan/phpstan/issues/3931

phpstan.neon.dist
```diff
@@ -1,2 +1,4 @@
 includes:
 	- phpstan-baseline.neon
+	- phpstan-docs-baseline.neon
+	- phpstan-unsupported-baseline.neon
```
AlexandruGG, line 4: - the docs baseline is just issues in `docs/` so nothing important but fixing them requires diving into other methods, which I will do anyway in future PRs.

- the unsupported baseline refers to the PHPStan unsupported feature I mentioned [here](https://github.com/loophp/collection/pull/96/files#r638333789). It's limited to the `unpack` operation and I expect we'll be able to get rid of it once `@template T of array` is supported

drupol: Nice and clean!

This is why I was using `@psalm-template`

By the way, what do you think about removing that `of array-key` ?
As this is a lazy library, the TKey can be of any type, so, I don't think I'm wrong by saying we can get rid of it. WDYT ?

src/Iterator/IterableIterator.php
drupol: Yes all good

src/Contract/Operation/Unpackable.php
drupol: Can't wait for this to be supported !

src/Collection.php
drupol: Should we report that as a bug ?

AlexandruGG: yeah leave it with me. Looks like a valid issue though maybe I thought it makes sense because it has no way to know where the template parameters are coming from

AlexandruGG: done! https://github.com/phpstan/phpstan/issues/5065

phpstan.neon.dist
AlexandruGG: I think you're right we could get rid of `array-key` but I don't think it can only be done in one place because it will probably result in static analysis errors.

I'll look to prepare a PR after this to remove it everywhere and see how many errors we get 😄

AlexandruGG:
@drupol btw what's the process for releases and how often do we do them in this project?
I can see the last one was on April 20 and there are quite a few things in master by now that could be released

drupol:
> @drupol btw what's the process for releases and how often do we do them in this project?
> I can see the last one was on April 20th and there are quite a few things in master by now that could be released

There is actually no strict release process and I'm absolutely not against to setup one.
Making a release is extremely easy, everything is automatically done by Github Actions.
The only thing that we need to do is generate the changelog with the docker command.

If you want, I can make a point release this morning.

Do you prefer to wait until this PR is merged or I do it before?

phpstan.neon.dist
drupol: Yeah I did it in #95 maybe it could be a starting point.

AlexandruGG:
> > @drupol btw what's the process for releases and how often do we do them in this project?
> > I can see the last one was on April 20th and there are quite a few things in master by now that could be released
> 
> There is actually no strict release process and I'm absolutely not against to setup one.
> Making a release is extremely easy, everything is automatically done by Github Actions.
> The only thing that we need to do is generate the changelog with the docker command.
> 
> If you want, I can make a point release this morning.
> 
> Do you prefer to wait until this PR is merged or I do it before?

This PR is ready from my point of view so let me know if there's anything left to address 😁 .
Would be great if you could do a release after this is merged! 👍

drupol (approved):

drupol:
4.0.7 released!

drupol:
Thanks again for your amazing contributions!

src/Contract/Operation/Unpackable.php
drupol: It's now supported! https://github.com/phpstan/phpstan-src/pull/673

AlexandruGG: nice job! 💯

PR #97, opened 2021-05-25
SA Checks: Methods Starting with "a"

AlexandruGG:
This PR initially aimed to add to the SA Checks all collection methods starting with the letter "a". 

After the discussions and the realisation that more `mixed` typing usage will be needed, I decided to shorten this to only the static analysis check for `all` + a few other minor changes.


src/Contract/Operation/Appendable.php
```diff
@@ -21,9 +21,11 @@ interface Appendable
     /**
      * Add one or more items to a collection.
      *
-     * @param mixed ...$items
+     * @template A
      *
-     * @psalm-return \loophp\collection\Collection<int|TKey, T>
+     * @param A ...$items
+     *
+     * @return \loophp\collection\Collection<int|TKey, A|T>
```
AlexandruGG, line 28: @drupol this is out of sync with the [Append operation](https://github.com/loophp/collection/blob/master/src/Operation/Append.php#L24), which I also want to update.

First I wanted to understand: is `append` meant to allow adding items of any type to the collection? Or only items of the type already contained? Example below:

```PHP

A) Collection::fromIterable([1, 2, 3])->append(4); // this is always fine
B) Collection::fromIterable([1, 2, 3])->append('bar'); // should this be allowed?
```

B) transforms the collection from `Collection<int, int>` into `Collection<int, string|int>`. I know this library supports this behaviour, but I'm not sure what your intention was when building it.

drupol: Yes, mixing types is definitely allowed.

tests/static-analysis/empty.php
```diff
@@ -14,15 +14,15 @@
 /**
  * @param Collection<int, int> $collection
  */
-function checkNumeric(Collection $collection): void
+function empty_checkNumeric(Collection $collection): void
```
AlexandruGG, line 17: I realised these functions need unique names because they're not under a namespace. I only got away with it so far because they had the same type annotations so the analysers didn't complain before 😄

src/Contract/Operation/Appendable.php
AlexandruGG: I see, in that case I think static analysis becomes less useful in many cases.
For `append` if it can be used like this `->append(4, 'foo', ['bar'])` then I guess `mixed` is the only type that can be used for the `$items`.

The problem is then that even if you do `A)` (append the same type) the static analysis will not be able to tell that the collection is still all of the same type.

AlexandruGG: I have mixed (pun intended 😄 ) feelings about this. On one hand I see the flexibility of mixing types but it's a bit inconvenient when the collection is used as a single-type collection because after many operations type hints will need to be used in the calling code to show that the type is still the same

drupol: I sharing the feelings.

However, this library has never been made to be a single-type collection.

I know that this would make typing more complex or even impossible, but this is how it is. I think we should keep that flexibility for the ease of use.

If we really want a strict collection, maybe we should use a `StrictCollection` class and do the stuff in it.

WDYT?

AlexandruGG: I would definitely like an option for a strictly-typed collection 👍 . Maybe we could call it `TypedCollection`.

How would it work with the type hints fo the operations though? Would the code effectively need to be duplicated but with different type hints?

drupol: Good question. I think this is part of a much broader discussion in its own branch, don't you think?

AlexandruGG: yep I agree, I might start one in https://github.com/loophp/collection/discussions at some point?

For this PR I'll update things to take into account collection usage with data of different types

drupol: A discussion is fine to start with, then a PR is also fine. There will be so much things to discuss that it's better to dedicate its own scope.

src/Operation/Append.php
```diff
@@ -14,28 +14,27 @@
...
- * @psalm-template TKey of array-key
- * @psalm-template T
+ * @template TKey
+ * @template T
  */
 final class Append extends AbstractOperation
 {
     /**
-     * @psalm-return Closure(T...): Closure(Iterator<TKey, T>): Iterator<int|TKey, T>
+     * @return Closure(mixed ...$items): Closure(Iterator<TKey, T>): Iterator<int|TKey, T|mixed>
```
AlexandruGG, line 23: the changes here bring it in line with the interface since items of different types can be appended

drupol: Are you sure about that? Shouldn't it be: 

```suggestion
     * @return Closure(...mixed): Closure(Iterator<TKey, T>): Iterator<int|TKey, T|mixed>
```

instead ?

AlexandruGG: Both are valid, I made a similar change here in previous [PR](https://github.com/loophp/collection/blob/master/src/Collection.php#L151). Also examples from the issue I submitted to PHPStan 😄 :
https://phpstan.org/r/474314ca-f343-4454-b100-dda6fb144bd2
https://psalm.dev/r/12f59e9b5e

The benefit of using the name of the parameter is that it makes it easier when reading to tie it to the $items [here](https://github.com/loophp/collection/blob/master/src/Operation/Append.php#L30)

AlexandruGG: Actually, the other version (`...mixed`) is not valid for PHPStan, only for Psalm. Whereas looks like this one is valid for both 🤷

drupol: Ho man, nice discovery! Going to use it everywhere !

src/Contract/Operation/Associateable.php
```diff
@@ -13,17 +13,19 @@
...
-     * @psalm-param null|callable(TKey, TKey, T, Iterator<TKey, T>):(T|TKey) $callbackForKeys
-     * @psalm-param null|callable(T, TKey, T, Iterator<TKey, T>):(T|TKey) $callbackForValues
+     * @template U
+     * @template V
      *
-     * @psalm-return \loophp\collection\Collection<TKey|T, T|TKey>
+     * @param null|callable(TKey, TKey, T, Iterator<TKey, T>): U $callbackForKeys
+     * @param null|callable(T, TKey, T, Iterator<TKey, T>): V $callbackForValues
+     *
+     * @return \loophp\collection\Collection<TKey|U, T|V>
```
AlexandruGG, line 28: the previous typing was more restrictive than what the operation actually allows you to do, because you can change the type of the keys and the values through the callbacks. However, this freedom is proving very difficult to type correctly and satisfy the analysers 😄

tests/static-analysis/associate.php
```diff
@@ -0,0 +1,54 @@
...
+    null,
+    static fn (int $carry, int $key, int $value) => $value
+)
+);
+associate_checkNumeric(Collection::fromIterable([4, 5, 6])->associate(
+    static fn (int $carry, int $key): int => $key * 2,
+    static fn (int $carry, int $key, int $value): int => $value * -1
+));
+
+associate_checkMap(Collection::fromIterable(['foo' => 1, 'bar' => 2])->associate(static fn (string $carry, string $key): string => "{$key}z"));
```
AlexandruGG, line 54: some of the errors here are ridiculous 😄 

`ERROR: InvalidArgument - tests/static-analysis/associate.php:54:20 - Argument 1 of associate_checkMap expects loophp\collection\Collection<string, int>, loophp\collection\Collection<non-empty-string, 1|2> provided (see https://psalm.dev/004)`

Can't tell if it's an issue with the type annotations or just pushing the limits of what Psalm is capable of right now

src/Contract/Operation/Associateable.php
drupol: Definitely.

tests/static-analysis/associate.php
drupol: It goes far indeed, but I don't know what's the best option here.

tests/static-analysis/associate.php
```diff
@@ -0,0 +1,54 @@
...
+associate_checkNumeric(
+    Collection::fromIterable([4, 5, 6])->associate(
+    null,
+    static fn (int $carry, int $key, int $value) => $value
+)
+);
+associate_checkNumeric(Collection::fromIterable([4, 5, 6])->associate(
+    static fn (int $carry, int $key): int => $key * 2,
+    static fn (int $carry, int $key, int $value): int => $value * -1
+));
```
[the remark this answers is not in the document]
drupol: Indeed. So weird!

AlexandruGG:
@drupol I think what I initially intended to do with these static-analysis checks is proving a bit infeasible 😓 .

I think I didn't realise the flexibility that the operations actually allow because I mostly looked at the type hints, which are more restrictive at the moment than what the library allows you to do. For example [Associateable](https://github.com/loophp/collection/blob/master/src/Contract/Operation/Associateable.php#L26) suggests that you will get back the same `TKey, T` but actually the types can be completely different depending on what transformation you have in the callbacks.

In addition, most of the examples in the [documentation](https://loophp-collection.readthedocs.io/en/latest/pages/api.html#methods-operations) are with collections of a single type.

I think that in order to allow for usage as intended, with a lot of flexibility, things need to be a bit more loosely-typed. That's probably one of the reasons why the [Laravel Collections](https://github.com/illuminate/collections/blob/master/Collection.php) typings are a lot less specific, i.e. more usage of `mixed` and `callable` but without specifying the types for keys and values. I'm not a fan of it but I don't see another way if the flexibility is the most important, because the specific type hints are too constraining at the moment.

drupol:
Thanks for being honest, I fully agree and I know that we will loose some typing information **because** of that flexibility. That's the trade off.

What do you think we should do in order to clean up the types in the library then?

AlexandruGG:
> > Thanks for being honest, I fully agree and I know that we will loose some typing information **because** of that flexibility. That's the trade off.
> > What do you think we should do in order to clean up the types in the library then?
> 
> I think we need to remove some of the template types and allow things to be less specific and `mixed`.
> But at the same time I also would want to keep the existing type hints (they must've taken a lot of work!) and tweak them as needed for the `TypedCollection` we were discussing in #99.
> 
> The best way I see this happening is in parallel: for example let's say I'm adding the static analysis checks for the `append` method. I would do it for both the regular Collection and the TypedCollection and ensure the static analysis passes for both. That way it would be easier to see the differences between them and ensure everything's working as intended.

@drupol what do you think about this approach? If you're okay with it I can start a new PR so you can see what it would look like, and for this PR I would remove most things apart from the `all` static analysis check and some other minor changes

drupol:
Hi Alex,

Sorry for the delay in my responses. I'm preparing my presentation at AFUP about this library, this is tomorrow: https://event.afup.org/afupday2021-interview-pol-dellaiera/ (_if you want to follow it, send me an email, I'll see what I can do!_)

Regarding your question, yes I agree obviously. As long as we won't use a typed collection, we will have to use `mixed` a bit everywhere.

docs/pages/code/e.php
```diff
@@ -15,9 +15,11 @@
 
 $addition = static fn (float $value1, float $value2): float => $value1 + $value2;
 
-$fact = static fn (float $number): float => (float) Collection::range(1, $number + 1)->foldLeft($multiplication, 1)->current();
+$fact = static fn (float $number): float => (float) Collection::range(1, $number + 1)
+    ->foldLeft($multiplication, 1)
+    ->current();
```
AlexandruGG, line 20: [This change](https://github.com/loophp/collection/commit/3f1855395c40675e8f9cf3eac070a2a82c867ab4) was causing a[ failure in CI](https://github.com/loophp/collection/runs/2687478640?check_suite_focus=true) and confusing me massively 😄

AlexandruGG:
No clue why unit tests are failing on macOS in CI but it's unrelated to the changes here 🤷‍♂️

docs/pages/code/e.php
drupol: Yeah sometimes those tools are extremely frustrating!

drupol:
Thanks mate! Now going to reply to your email :)

PR #101, opened 2021-05-28
Update Psalm Annotations

AlexandruGG:
This PR is just #95 with fixes for the code analysis.

- [x] Replace @psalm-template with @template.
- [x]  Replace @psalm-param with @param.
- [x]  Replace @psalm-var with @var.
- [x]  Replace @psalm-return with @return.


src/Collection.php
```diff
@@ -367,7 +367,7 @@ public function frequency(): CollectionInterface
      * @template NewTKey
      * @template NewT
      *
-     * @param callable(...mixed): Iterator<NewTKey, NewT> $callable
+     * @param callable(mixed ...$parameters): iterable<NewTKey, NewT> $callable
```
AlexandruGG, line 374: - the `...mixed` syntax is not supported by PHPStan, but the `mixed ...$parameters` is supported by both Psalm and PHPStan

- also note the collection can be instantiated with a callable returning any `iterable`, not just Iterator, as exemplified in the [SA check](https://github.com/loophp/collection/blob/master/tests/static-analysis/fromCallable.php#L144-L146)

drupol: Great!

AlexandruGG:
Any clue what we could do about the unit tests failing on macOS? Have you seen this phpspec error before?
```
collaborator does not exist : spec\loophp\collection\Iterator\CacheItemPoolInterface

drupol:
> Any clue what we could do about the unit tests failing on macOS? Have you seen this phpspec error before?
> 
> ```
> collaborator does not exist : spec\loophp\collection\Iterator\CacheItemPoolInterface
> ```

This is temporary, I guess it will be fixed upstream very soon. I guess we can agree to ignore it for now. ok ?

AlexandruGG:
> > Any clue what we could do about the unit tests failing on macOS? Have you seen this phpspec error before?
> > ```
> > collaborator does not exist : spec\loophp\collection\Iterator\CacheItemPoolInterface
> > ```
> 
> This is temporary, I guess it will be fixed upstream very soon. I guess we can agree to ignore it for now. ok ?

Yeah that's fine. This is ready otherwise, but take your time to review again as it's quite big 😄

drupol (approved):

PR #102, opened 2021-05-30
Improve Collection Typings + Add `strict` Operation

AlexandruGG:
This PR starts a process of improving the types on the operations to make the Collection behave like a single-type one, without enforcing the type restriction at runtime unless the `strict` operation is used. Based on multiple discussions had in this PR and in #99.


tests/static-analysis/CollectionT/append.php
```diff
@@ -0,0 +1,97 @@
...
+ * @phpstan-ignore-next-line
+ */
+appendT_checkMap(CollectionT::fromIterable($foo)->append(3));
+
+/** @psalm-suppress InvalidArgument @phpstan-ignore-next-line */
+appendT_checkMixed(CollectionT::empty()->append(1, [3]));
+/** @psalm-suppress InvalidArgument */
+appendT_checkMixed(CollectionT::empty()->append(...[1, [3]]));
+/** @psalm-suppress InvalidArgument @phpstan-ignore-next-line */
+appendT_checkMixed(CollectionT::fromIterable(['foo' => 1, 'bar' => '2'])->append(1, ['3']));
```
AlexandruGG, line 97: @drupol have a look at this file comparing with the [other one](https://github.com/loophp/collection/pull/102/files?diff=split&w=1#diff-735f11d2c95768f8eed8f60feffd52f4225859f05c6a9c526695ef30108fbf23R1) (mixed collection) to understand the difference in what `CollectionT`'s interface would allow. Even though it doesn't yet have any runtime difference (doesn't enforce a single type), static analysis would stop the user from adding mixed types to it. Also note the difference in the `@var` annotations needed (or not needed for the typed version).

tests/static-analysis/CollectionT/fromIterable.php
```diff
@@ -0,0 +1,63 @@
...
+$arrayIteratorMap = new ArrayIterator(['foo' => 1, 'bar' => 2]);
+$arrayIteratorMixed = new ArrayIterator([1, 2, '3', 'b', 5]);
+
+fromIterableT_checkList(CollectionT::fromIterable($generatorList()));
+fromIterableT_checkMap(CollectionT::fromIterable($generatorMap()));
+fromIterableT_checkMixed(CollectionT::fromIterable($generatorMixed()));
+
+fromIterableT_checkList(CollectionT::fromIterable($arrayList));
+fromIterableT_checkMap(CollectionT::fromIterable($arrayMap));
+fromIterableT_checkMixed(CollectionT::fromIterable($arrayMixed));
```
AlexandruGG, line 59: when it comes to instantiation however, static analysis cannot "protect" the user of `CollectionT`, or at least I haven't found a way to do it. Because the template type `T` is interpreted by static analysers as a union type like `int|string`, which is still valid in PHP

AlexandruGG:
As an aside, there seem to be some differences between the CI setup and running GrumPHP locally. The failures seen in CI don't happen on local. Do you think it could be due to the fact that Psalm is configured to [use a cache in CI](https://github.com/drupol/php-conventions/blob/master/config/php73/grumphp.yml#L107)? (I think we should disable this btw)

```
ERROR: UnusedPsalmSuppress - tests/static-analysis/CollectionT/append.php:79:21 - This suppression is never used (see https://psalm.dev/207)
/** @psalm-suppress InvalidScalarArgument @phpstan-ignore-next-line */


ERROR: UnusedPsalmSuppress - tests/static-analysis/CollectionT/append.php:81:21 - This suppression is never used (see https://psalm.dev/207)
/** @psalm-suppress InvalidScalarArgument @phpstan-ignore-next-line */


ERROR: UnusedPsalmSuppress - tests/static-analysis/CollectionT/append.php:96:21 - This suppression is never used (see https://psalm.dev/207)
/** @psalm-suppress InvalidArgument @phpstan-ignore-next-line */```

File: tests/static-analysis/CollectionT/append.php (hunk 1 above)
Comment (drupol) [reply tests/static-analysis/CollectionT/append.php:97]: This is beautiful, it should be the default!

File: tests/static-analysis/CollectionT/fromIterable.php (hunk 2 above)
Comment (drupol) [reply tests/static-analysis/CollectionT/fromIterable.php:59]: Mmmh mmh I see. I don't have the answer right now to this.

**Comment by drupol** [comment]
> As an aside, there seem to be some differences between the CI setup and running GrumPHP locally. The failures seen in CI don't happen on local. Do you think it could be due to the fact that Psalm is configured to [use a cache in CI](https://github.com/drupol/php-conventions/blob/master/config/php73/grumphp.yml#L107)? (I think we should disable this btw)

No I don't think this is the case.

However, you can try to disable it by adding:

`tasks.psalm.no_cache: true`

in the local grumphp file, under the `parameters` directive.

File: tests/static-analysis/CollectionT/append.php (hunk 1 above)
Comment (drupol) [reply tests/static-analysis/CollectionT/append.php:97]: So basically your idea for making this happening is to not enforce any type? Just enforce the type through the annotations?
The consequence of this will be a change in operations, making them non-variadic. Right?

Comment (AlexandruGG) [reply tests/static-analysis/CollectionT/append.php:97]: I think your **"Idea 1"** from the discussion https://github.com/loophp/collection/discussions/99#discussion-3381462 is still valid and would be useful. It should be used _in conjunction_ with the type annotations, but it might be limited due to what `gettype` can do, unless we use it recursively and check all the keys and value types in the case where you have a collection of arrays.

Also I am not sure about **union types** because static analysis will allow instantiation, as mentioned [here](https://github.com/loophp/collection/pull/102#discussion_r642100624), but the runtime enforcement solely based on `gettype` will throw an exception. Perhaps it wouldn't be an issue though as long as the user knows about the behaviour.

Regarding the **operations** I don't think there would be any implications in how they work, they would still be variadic. See the examples in the static analysis checks for `append` - it still allows the user to append multiple elements; the only inconvenience might be that _sometimes_ extra manual type hints might be needed on the user side to satisfy the static analysis. But this will also be dependant on the strictness levels they use for psalm, phpstan

File: tests/static-analysis/CollectionT/fromIterable.php (hunk 2 above)
Comment (AlexandruGG) [reply tests/static-analysis/CollectionT/fromIterable.php:59]: in an _eager_ collection there would typically be an instantiation check which loops through the elements, checks the type, and throws an exception if they are mixed, but ofc in a lazy one your Idea 1 is the way to go. I think it should be fine as long as it's documented

**Comment by AlexandruGG** [comment]
> > As an aside, there seem to be some differences between the CI setup and running GrumPHP locally. The failures seen in CI don't happen on local. Do you think it could be due to the fact that Psalm is configured to [use a cache in CI](https://github.com/drupol/php-conventions/blob/master/config/php73/grumphp.yml#L107)? (I think we should disable this btw)
> 
> No I don't think this is the case.
> 
> However, you can try to disable it by adding:
> 
> `tasks.psalm.no_cache: true`
> 
> in the local grumphp file, under the `parameters` directive.

You're right looks like disabling the cache didn't fix the issue. I can't explain the difference in the errors reported but I've reverted the Psalm config change to report unused psalm-suppress annotations which was causing the issue. I'll probably just use it in local development.

File: tests/static-analysis/CollectionT/append.php (hunk 1 above)
Comment (drupol) [reply tests/static-analysis/CollectionT/append.php:97]: Well, the idea is to keep this library simple, well documented. If we start shipping 2 collection classes, it will confuse people.

What I think is the following:

We should update `Collection.php` and use proper annotations, as if it was a typed collection, without enforcing it.

Then we could provide a new operation (**Idea 2** from #99) if we want to enforce it.

About the `gettype`, I don't think this is the valid way. Let's say you want to enforce a particular type of object (_implementing a specific interface_), when using `gettype` only, it's simply not possible because it will always return 'object'. 

WDYT?

Comment (AlexandruGG) [reply tests/static-analysis/CollectionT/append.php:97]: > We should update Collection.php and use proper annotations, as if it was a typed collection, without enforcing it.

I would like that but I thought we established it's not possible if we want to allow for the flexibility to mix types. And because of this 2x different interfaces are needed for the operations (at least for some of them, maybe not for all). Let's look at `Appendable`:

A) Typed:
```PHP
/**
 * @template TKey
 * @template T
 */
interface AppendableT
{
    /**
     * Add one or more items to a collection.
     *
     * @template U of T
     *
     * @param U ...$items
     *
     * @return CollectionT<int|TKey, T|U>
     */
    public function append(...$items): CollectionT;
```

B) Mixed:
```PHP
/**
 * @template TKey
 * @template T
 */
interface Appendable
{
    /**
     * Add one or more items to a collection.
     *
     * @param mixed ...$items
     *
     * @return Collection<int|TKey, T|mixed>
     */
    public function append(...$items): Collection;
}
```

Both are valid and have their own use cases. A) only allows you to append elements of the same type that you already have in the collection, whereas B) allows you to append anything you want.

If you like the idea of having a single collection class more then would you be happy to restrict the way it can be used?

drupol: I understand for sure.

So to summarize, in the end, we will end up with 2 classes (`Collection` and `CollectionT` for example) with exactly the same PHP code, but with different annotation?

AlexandruGG: actually looking at this PR and trying out a few things it looks like we won't need to duplicate the code for the operations, only the interfaces would have to be different.

So if this is indeed correct then to summarise we would have:

`Collection`
- implements `src/Contract/Collection`, which extends all the current operation interfaces but modified a bit to allow for the mixed usage without code analysis errors
- uses the operations in `src/Operation` just like it does now

`CollectionT`
- implements `src/Contract/CollectionT`, which extends new operation interfaces `AppendableT`, `ApplyableT` with updated type hints
- uses the same operations in `src/Operation`
- enforces basic types at runtime, i.e. primitives like `int`, `string`, `array`. For objects we can use `get_class` to ensure the same class. I think this is important for the end user because it justifies a bit more choosing which collection to use. When you _know_ you want the same type you don't want to solely rely on static analysis because that can be worked around/ignored

In essence the only code that would be actually duplicated would be the methods themselves in `src/Collection.php` and `src/CollectionT.php`. One thought I had was to put them on an abstract class but due to requiring different interfaces it's not possible

AlexandruGG: @drupol hey, sorry I forgot to tag you here 😄. What do you think, are you on board with this?
I could add the runtime type enforcing iterator to this PR and it can be the start of CollectionT

drupol: > One thought I had was to put them on an abstract class but due to requiring different interfaces it's not possible.

Do you think we could have the same interface for both type of Collection? Wouldn't it be better?

AlexandruGG: I'm a bit confused 😓. Are you saying we should update the Collection interface to not allow mixing types? Because if we are to maintain the current flexibility it makes sense to have two separate interfaces like I described [here](https://github.com/loophp/collection/pull/102#discussion_r642635206).

drupol: Ok... I don't like the idea of duplicating each interfaces. I'm thinking about something simpler.
I guess that as long as we will have 2 Collections, we must have twice all the interfaces. 
Do you see a better solution?

AlexandruGG: I see. I think the separate interfaces would property show the _intent_ behind each collection. 

The only simpler solution that I can think of if we want to avoid this is: treat `Collection` like a `CollectionT` in the type system, but not at runtime. It would have to be specified in the documentation and when you'd want to use it like a mixed collection you might sometimes need to use annotations to satisfy the static analysis. I think it's not a bad idea, in this case we can ditch CollectionT altogether and just have the `strict` operation you mentioned in Idea 2

drupol: Yes, this is is exactly what I have in mind. If you're ok with that, we can go on. Deal ? :)

AlexandruGG: alright, let's give it a shot 😁. I'll refactor this PR to get rid of CollectionT and start some work towards the plan

drupol: Excellent. Let me know if I can be of any help on this.

AlexandruGG: I do have a question for you actually. What would you say about this implementation of `strict` (note: haven't tested yet, just something that suddenly came to mind):

```PHP
public function strict(): CollectionInterface 
{
    return self::fromCallable(static fn() => yield from (new TypedIterator($this->getIterator()));
}
```

This way we don't need a new operation and the logic is in the `TypedIterator` like in _Idea 1_. And the user doesn't need to provide any callback to `strict` because the logic is internal.

drupol: This seems to be at first sight a possible solution. I'll test tomorrow, I'm about to sleep here.

src/Iterator/TypedIterator.php
```diff
@@ -0,0 +1,61 @@
...
+                        );
+                    }
+
+                    yield $key => $value;
+                }
+            },
+            $iterable
+        );
+    }
+}
```
AlexandruGG, line 61: @drupol added this now based on Idea 1, with a few minor modifications. Please see the [spec](https://github.com/loophp/collection/pull/102/files#diff-34c434dbd5917f3fc5a051c137420d165761f1eef3805d8ba9abbbc49fbbf9b2R18) to understand the complete behaviour.

DONE:
- [x] specs for all scalars: `int, float, string, bool`
- [x] allow `NULL` in addition to another type: I thought this is best because having nullable types is quite common in php; I think it would be too restrictive to _not_ allow it

TODO:
- [ ] implement logic and tests for other types: `resource, array, object`. For object check class, for resource check to not have a mix of closed and not closed, and for array I think don't check inside for now
- [ ] use `TypedIterator` in the `strict` operation like I described [here](https://github.com/loophp/collection/pull/102#discussion_r646189027)

drupol: Hi!

This morning I updated it, let me know what you think: https://gist.github.com/drupol/15ddab9628e4cd2f90321ad2f0368bd9

AlexandruGG: Morning, that's interesting so you want to allow user the user the option to define how the typing is enforced. Sounds good to me :).

Another point:
- I think `yield` should happen after the type checking. If you try to run the tests with your version you will see that they only pass on the call to `next` after you passed the element which was of a different type

I will use the gist to update this when I get the chance and add more tests, thanks!

drupol: > Morning, that's interesting so you want to allow user the user the option to define how the typing is enforced. Sounds good to me :).

Yes, there is a reason to that.
This `getType` is a kind of basic implementation, it will never cover all the user needs. Then, in those specific case, the user is able to customize that `callable`.
If you checkout the operations, I often let the user customize the stuff with callables, so it's fully flexible. I'm trying to never define/set something in stone within an operation.

> Another point:
> 
>     * I think `yield` should happen after the type checking. If you try to run the tests with your version you will see that they only pass on the call to `next` after you passed the element which was of a different type

Mmmh but this is what I did, the `yield` statement is after the type checks.

> I will use the gist to update this when I get the chance and add more tests, thanks!

Sure, let's see what comes out.

AlexandruGG: > Mmmh but this is what I did, the yield statement is after the type checks.

Oh you're right, it only yields early if the value is `null` 👍

src/Collection.php
```diff
@@ -714,6 +716,11 @@ public function squash(): CollectionInterface
         return self::fromIterable($this->pack()->all())->unpack();
     }
 
+    public function strict(?callable $callback = null): CollectionInterface
+    {
+        return self::fromCallable(fn (): Generator => yield from (new TypedIterator($this->getIterator(), $callback)));
+    }
```
AlexandruGG, line 722: I updated the PR with your gist and more tests, and added the `strict` operation now. I think this is nice, however, I'm wondering whether we could have an alternative so if I'm using the collection like a strict one I don't have to do `Collection::fromIterable([1,2])->strict()` all the time.

What if there was an optional boolean which was stored on a property on the class and based on it `getIterator` would use the Typed one or not. For example:


```PHP
public static function fromIterable(iterable $iterable, bool $strict = false): self
{
   	return new self(
        static fn (iterable $iterable): Iterator => new IterableIterator($iterable),
        $strict,
        $iterable
    );
}

public function getIterator(): Iterator
{
    $closureIterator = new ClosureIterator($this->source, ...$this->parameters);
	
    return $this->strict ? new TypedIterator($closureIterator): $closureIterator;
}


Collection::fromIterable([1, 2, 3], strict: true);
```

This would mean that the type enforcement is maintained throughout operation chaining, because of the way operations use the `getIterator` method.

Otherwise, the user would always have to remember to call `strict` after a chain of operations.

Does this sound like a crazy idea or does it make sense to you? 😄  @drupol

drupol:
Hi Alex,

I reviewed locally your pull request. I can't wait to merge this :)

However I have a few remarks:

* I think it would be better in this case to create an `Operation` instead of the `TypedIterator`. The `strict` should follow the same pattern as the other operations, at least for version `^4`.
* I understand the feeling of having a "already made" typed collection class, but I think we should not. If we want to achieve such things, we should add the documentation and explain how to create it's own class `TypedCollection` extending `Collection`.
I actually do not like the idea of starting to pass parameters in constructors. `Collection` should stay very basic and flexible so people can implements whatever they want. 

WDYT?

src/Collection.php
drupol: I replied in the thread down there !

AlexandruGG:
> Hi Alex,
> 
> I reviewed locally your pull request. I can't wait to merge this :)
> 
> However I have a few remarks:
> 
> * I think it would be better in this case to create an `Operation` instead of the `TypedIterator`. The `strict` should follow the same pattern as the other operations, at least for version `^4`.
> * I understand the feeling of having a "already made" typed collection class, but I think we should not. If we want to achieve such things, we should add the documentation and explain how to create it's own class `TypedCollection` extending `Collection`.
>   I actually do not like the idea of starting to pass parameters in constructors. `Collection` should stay very basic and flexible so people can implements whatever they want.
> 
> WDYT?

Hey @drupol , thanks for this.

- regarding the `Operation` instead of `TypedIterator` I can try to refactor this. Just to clarify, this means that we still use the `TypedIterator` but it gets moved to an operation, similar to how the `RandomIterator` is used in [Shuffle](https://github.com/loophp/collection/blob/master/src/Operation/Shuffle.php#L37), right?

- I understand the concern regarding passing parameters. I'm personally not a fan of allowing `Collection` to be extendable, I think it's best to keep it `final` like it is now though. I was going to suggest providing a `TypedCollection` which is very small, just extending `Collection` and overriding `getIterator`. However, that's not possible if we want to keep it final. I think for now let's go with the `strict` operation.

drupol:
>     * regarding the `Operation` instead of `TypedIterator` I can try to refactor this. Just to clarify, this means that we still use the `TypedIterator` but it gets moved to an operation, similar to how the `RandomIterator` is used in [Shuffle](https://github.com/loophp/collection/blob/master/src/Operation/Shuffle.php#L37), right?

Yes, this is also a valid option.
 
>     * I understand the concern regarding passing parameters. I'm personally not a fan of allowing `Collection` to be extendable, I think it's best to keep it `final` like it is now though. I was going to suggest providing a `TypedCollection` which is very small, just extending `Collection` and overriding `getIterator`. However, that's not possible if we want to keep it final. I think for now let's go with the `strict` operation.

Wait wait wait, it's not because you have the `final` keyword that its not possible to create another Collection class that extends the functionalities of the original one.

I'm all in favor of using the `Composition` pattern instead of `Inheritance`, I found it so much easier and cleaner in every way.

I can provide an example if you want to, but a bit later today.

drupol:
Here's an extremely basic (and streamlined) example on how to extend the collection: https://gist.github.com/drupol/8b646ec5e1b91d45cf20969b4eabb3af

AlexandruGG:
> Here's an extremely basic (and streamlined) example on how to extend the collection: https://gist.github.com/drupol/8b646ec5e1b91d45cf20969b4eabb3af

I see what you mean. When you said "extend collection" I interpreted it literally as `extends` aka inheritance 😄.

This is nice from a composition point of view, however I'm not sure it's very convenient. All of the methods from Collection would have to be repeated in this class, right?

drupol:
> > Here's an extremely basic (and streamlined) example on how to extend the collection: https://gist.github.com/drupol/8b646ec5e1b91d45cf20969b4eabb3af
> 
> I see what you mean. When you said "extend collection" I interpreted it literally as `extends` aka inheritance 😄.
> 
> This is nice from a composition point of view, however I'm not sure it's very convenient. All of the methods from Collection would have to be repeated in this class, right?

Yes, this is up to the user to implement it.

All of the method needs be reimplemented by the user as well.

There are so much methods that it could be indeed cumbersome... But it is how it is when we use composition.

docs/conf.py
```diff
@@ -57,9 +57,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = u'1.0.0'
+version = u'4.0.0'
 # The full version, including alpha/beta/rc tags.
-release = u'1.0.0'
+release = u'4.0.0'
```
AlexandruGG, line 62: I think these affect the version that appears here:

Screenshot 2021-06-11 at 23 35 53

This should be fairly in sync with the releases otherwise it might lead someone to believe that the documentation is out of date

README.md
```diff
@@ -13,7 +13,7 @@
 
 ## Description
 
-Collection is a functional utility library for PHP greater than 7.4.
+Collection is a functional utility library for PHP greater than 7.4, including PHP 8.0.
```
AlexandruGG, line 16: I think it's good to mention this to reassure users that they can use it in PHP 8

AlexandruGG:
@drupol refactored into a `Strict` operation now + added documentation.

There's an issue with phpcsfixer and phpcs where they are in opposition -> phpcs wants space here `new class () {}` and phpcsfixer wants no space, like `new class() {}` 😄

drupol:
> @drupol refactored into a `Strict` operation now + added documentation.
> 
> There's an issue with phpcsfixer and phpcs where they are in opposition -> phpcs wants space here `new class () {}` and phpcsfixer wants no space, like `new class() {}` 😄 

I think you can get rid of the parenthesis after the class keyword ?

README.md
drupol: Sure it's good

docs/conf.py
drupol: Defo.

AlexandruGG:
> > @drupol refactored into a `Strict` operation now + added documentation.
> > There's an issue with phpcsfixer and phpcs where they are in opposition -> phpcs wants space here `new class () {}` and phpcsfixer wants no space, like `new class() {}` 😄
> 
> I think you can get rid of the parenthesis after the class keyword ?

I tried but you cannot, because phpcsfixer adds it back 😄

AlexandruGG:
looks like this issue: https://github.com/squizlabs/PHP_CodeSniffer/issues/3200

drupol:
It has also been reported here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5463

drupol:
I think you should fix it by doing this: https://gist.github.com/drupol/39a459e29764efdc41cfacb28ee4b751

drupol:
I added a note on how to extend this library here: https://loophp-collection.readthedocs.io/en/latest/pages/usage.html#extending-collection

AlexandruGG:
> I added a note on how to extend this library here: https://loophp-collection.readthedocs.io/en/latest/pages/usage.html#extending-collection

Nice 👍

AlexandruGG:
> I think you should fix it by doing this: https://gist.github.com/drupol/39a459e29764efdc41cfacb28ee4b751

do you not think it's better to just ignore the `docs` folder? PHPCSFIXER is running on that anyway and I think it's nicer than seeing the phpcs ignore comments in the docs

AlexandruGG:
> I added a note on how to extend this library here: https://loophp-collection.readthedocs.io/en/latest/pages/usage.html#extending-collection

This also broke PHPStan but I've added an ignore for `docs` now

drupol:
> > I think you should fix it by doing this: https://gist.github.com/drupol/39a459e29764efdc41cfacb28ee4b751
> 
> do you not think it's better to just ignore the `docs` folder? PHPCSFIXER is running on that anyway and I think it's nicer than seeing the phpcs ignore comments in the docs

Ok!

AlexandruGG:
@drupol let's get this released if you've had the chance to review 🚀

drupol:
Thanks mate! Very nice stuff :)

AlexandruGG:
> Thanks mate! Very nice stuff :)

Thank you for the help on this, I think we had some good collaboration!

PR #104, opened 2021-06-01
PHP 8 Compatibility

AlexandruGG:
This PR ensures this library is always checked for full PHP 8 compatibility, based on the discussion https://github.com/loophp/collection/discussions/103

- [x] updating CI workflows to run on PHP 8 as well as PHP 7.4
- [x] fixing a couple static analysis issues
- [x] handle usage of `asyncMap` in PHP in the `CollectionSpec`. Due to an issue https://github.com/phpspec/phpspec/issues/1375 with PHPSpec the test fails. Should be fixed soon however
- [x] requiring the PHP 8 jobs to merge in the `settings.yml`


src/Contract/Operation/Applyable.php
```diff
@@ -20,9 +21,9 @@ interface Applyable
     /**
      * Execute a callback for each element of the collection.
      *
-     * @param callable(TKey, T):bool ...$callables
+     * @param callable(T, TKey, Iterator<TKey, T>): bool ...$callbacks
```
AlexandruGG, line 24: the interface was out of sync with the [operation](https://github.com/loophp/collection/pull/104/files#diff-79d43bca592b76fa0fb2583de4ea453a994e67ea88ab8d27e403579525c8c834R31)

spec/loophp/collection/CollectionSpec.php
```diff
@@ -226,9 +230,18 @@ public function it_can_asyncMap(): void
...
-            ->shouldIterateAs(['a' => 2, 'b' => 4, 'c' => 6]);
+        $this->beConstructedThrough('fromIterable', [['c' => 3, 'b' => 2, 'a' => 1]]);
+
+        if (PHP_VERSION_ID >= self::PHP_VERSION_8) {
+            $this
+                ->asyncMap($callback1, $callback2)
+                ->shouldThrow(Error::class)
+                ->during('all');
+        } else {
+            $this
```
AlexandruGG, line 241: I've modified this test case to show that an `Error` is thrown for `asyncMap` in PHP > 8.0.

That way the test still passes and we have this visibility

.github/settings.yml
```diff
@@ -12,13 +12,21 @@ branches:
               strict: true
               contexts:
                   - "Static Analysis (ubuntu-latest, 7.4)"
+                  - "Static Analysis (ubuntu-latest, 8.0)"
                   - "Static Analysis (windows-latest, 7.4)"
+                  - "Static Analysis (windows-latest, 8.0)"
```
AlexandruGG, line 17: I assume these "required to merge" settings will only apply for future PRs after this is merged, right?

src/Contract/Operation/Applyable.php
drupol: Maybe we should fix that in another PR, wdyt?

spec/loophp/collection/CollectionSpec.php
drupol: I did [this](https://github.com/friends-of-phpspec/phpspec-requires) once upon a time, maybe we should see if it would be a better choice?

.github/settings.yml
drupol: yes :)

spec/loophp/collection/CollectionSpec.php
AlexandruGG: that would be nice but even better actually I was going to ask if we should just throw an intentional exception from `asyncMap` itself if it's called in an environment running php 8. What do you think? @drupol 

That way we make it clear it's unsupported for now and we can add this info the [documentation](https://loophp-collection.readthedocs.io/en/latest/pages/api.html#asyncmap) as well

drupol: Yes this is a good idea.

src/Contract/Operation/Applyable.php
AlexandruGG: well it's causing an issue with PHPStan in this PR so that's why I updated it here 😄 

It's related to [this error which was ignored](https://github.com/loophp/collection/pull/104/files#diff-22fe68f6ec4412b512a01a6526b11f8ad2f9e9c07ebcd32d018a131037059de7L14-L16). In PHP 8 the parameter name for `mb_strtolower` was changed to from `$str` to `$string`, so there would be x2 errors now depending on which PHP version PHPStan is run. I thought I'd just go ahead and fix it :)

spec/loophp/collection/CollectionSpec.php
AlexandruGG: updated 😁

drupol: Saw it, seems like it is what we need. Nice!

src/Operation/AsyncMap.php
```diff
@@ -41,6 +44,10 @@ final class AsyncMap extends AbstractOperation
      */
     public function __invoke(): Closure
     {
+        if (PHP_VERSION_ID >= Constant::PHP_VERSION_8) {
+            throw new RuntimeException('This operation is currently not supported on PHP 8.');
+        }
```
AlexandruGG, line 49: this needs to be inside the class otherwise PHPStan fails when discovering the files and that error cannot be ignored. Won't make a difference for the end user

.scrutinizer.yml
```diff
@@ -1,12 +1,19 @@
...
             tests:
                 override:
                     - php-scrutinizer-run
+        php80:
+            environment:
+                php:
+                    version: 8.0
+            tests:
+                override:
+                    - php-scrutinizer-run
```
AlexandruGG, line 16: this is based on [their docs](https://scrutinizer-ci.com/docs/build/multiple_test_environment#use-case-testing-different-versions), but I'm not familiar with this tool at all. I think this will only apply after this PR is merged, right?

src/Contract/Operation/Applyable.php
drupol: Ok.

.scrutinizer.yml
drupol: Yes :)

src/Contract/Operation/Applyable.php
drupol: I made a commit on `master` to fix it, you can rebase your branch if you want.

composer.json
```diff
@@ -20,7 +20,7 @@
         }
     ],
     "require": {
-        "php": ">= 7.4"
+        "php": "^7.4 || ^8.0"
```
drupol, line 23: Why this change ?

See: https://github.com/symfony/symfony/pull/36876

src/Collection.php
```diff
@@ -166,9 +166,9 @@ public function append(...$items): CollectionInterface
         return new self(Append::of()(...$items), $this->getIterator());
     }
 
-    public function apply(callable ...$callables): CollectionInterface
+    public function apply(callable ...$callbacks): CollectionInterface
```
drupol, line 169: I think we should normalize this everywhere.

Let's leave it like this today, but this is something I would like to change and normalize everywhere.

drupol (changes requested):
Hi @AlexandruGG,

Very nice work, thanks!

I gave my feedback, let me know what you think.

composer.json
AlexandruGG: Lots of heated discussions in that PR 😄 .

My main rationale is the one here: https://github.com/symfony/symfony/pull/36876#issuecomment-631351947. I don't see any good arguments against that comment in the thread. Basically we don't know for example how this library will perform in PHP 8.1, so if we use >= 7.4 then a user installing it on 8.1 might have a bad experience if they find that it's not working.

That being said I don't feel too strongly about this, because there's an argument to be made that users might report issues so you can thus fix them earlier.

AlexandruGG:
Don't merge this yet as I want to see what comes out of the discussion on `asyncMap` https://github.com/loophp/collection/discussions/103. It looks like the library is working in PHP 8, it's just PHPSpec that's failing

AlexandruGG:
@drupol hey, I think we should merge this because it adds the PHP 8 CI jobs and so will ensure any future code changes will be compatible with PHP 8.

It might take a bit to get the phpspec [issue](https://github.com/phpspec/phpspec/issues/1375) merged and released, but at that point we can just update [this test](https://github.com/loophp/collection/pull/104/files#diff-3682f293747f0968dc35d26e4ed1f8d6d059c378d0b81f1100f3dda60c09769cR233-R237).

PR #105, opened 2021-06-12
Add When operation.

drupol:
This PR

* [x] Add When operation
* [x] Has documentation
* [x] Has tests

### DIFF
--- a/src/Contract/Operation/Whenable.php
+++ b/src/Contract/Operation/Whenable.php
@@ -1,0 +1,29 @@
+<?php
+
+/**
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace loophp\collection\Contract\Operation;
+
+use Iterator;
+use loophp\collection\Contract\Collection;
+
+/**
+ * @template TKey
+ * @template T
+ */
+interface Whenable
+{
+    /**
+     * @param callable(Iterator<TKey, T>): bool $predicate
+     * @param callable(Iterator<TKey, T>): iterable<TKey, T> $whenTrue
+     * @param callable(Iterator<TKey, T>): iterable<TKey, T> $whenFalse
+     *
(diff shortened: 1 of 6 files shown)


AlexandruGG:
At first glance I can't tell what's the difference between this and `IfThenElse`? 🤔

drupol:
> At first glance I can't tell what's the difference between this and `IfThenElse`? thinking

This operation operates on the collection object once, while the `ifThenElse` operates on each item of the collection (this is an alias for a map).

src/Contract/Operation/Whenable.php
```diff
@@ -0,0 +1,29 @@
...
+interface Whenable
+{
+    /**
+     * @param callable(Iterator<TKey, T>): bool $predicate
+     * @param callable(Iterator<TKey, T>): iterable<TKey, T> $whenTrue
+     * @param callable(Iterator<TKey, T>): iterable<TKey, T> $whenFalse
+     *
+     * @return \loophp\collection\Collection<TKey, T>
+     */
+    public function when(callable $predicate, callable $whenTrue, ?callable $whenFalse = null): Collection;
```
AlexandruGG, line 28: should this return `Collection` the implementation or `Collection` the contract?

Isn't there a mismatch here? the `@return` is showing the implementation, but the actual return type of the function is the contract

AlexandruGG:
> > At first glance I can't tell what's the difference between this and `IfThenElse`? thinking
> 
> This operation operates on the collection object once, while the `ifThenElse` operates on each item of the collection (this is an alias for a map).

I understand now, thanks for clarifying this in the docs! 🙌

AlexandruGG:
@drupol do you mind adding the [static analysis check](https://github.com/loophp/collection/tree/master/tests/static-analysis) with any new operation please? This ensures we get the types right on the user side 😁

drupol:
True ! I have to get used to that, will do that later in the day for sure.

PR #106, opened 2021-06-12
Static Analysis Checks: Common Operations (I)

AlexandruGG:

- [x] apply
- [x] chunk
- [x] contains
- [x] count
- [x] current

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -450,8 +450,11 @@ Signature: ``Collection::compact(...$values);``
 contains
 ~~~~~~~~
 
-Check if the collection contains one or more values. Note that if multiple values are passed the operation
-acts like a logical ``OR``.
+Check if the collection contains one or more values.
+
+.. warning:: The `values` parameter is variadic and will be evaluated as a logical ``OR``.
+             If you're looking for a logical ``AND``, you have to make separate calls to this method;
+             note the calls cannot be in succession because the collection will contain a boolean after the first call.
 
 Interface: `Containsable`_
 
@@ -580,7 +583,7 @@ dropWhile
 Iterate over the collection items and takes from it its elements from the moment when the condition fails for the
 first time till the end of the list.
 
-.. warning:: The `callbacks` parameter is variadic and they are evaluated as a logical ``OR``.
+.. warning:: The `callbacks` parameter is variadic and will be evaluated as a logical ``OR``.
              If you're looking for a logical ``AND``, you have to make multiple calls to the
              same operation.
 
@@ -638,7 +641,7 @@ every
 
 This operation tests whether all elements in the collection pass the test implemented by the provided callback(s).
 
-.. warning:: The `callbacks` parameter is variadic and they are evaluated as a logical ``OR``.
+.. warning:: The `callbacks` parameter is variadic and will be evaluated as a logical ``OR``.
              If you're looking for a logical ``AND``, you have to make multiple calls to the
              same operation.
 
@@ -962,7 +965,7 @@ has
 
 Check if the collection has values.
 
-.. warning:: The `callbacks` parameter is variadic and they are evaluated as a logical ``OR``.
+.. warning:: The `callbacks` parameter is variadic and will be evaluated as a logical ``OR``.
              If you're looking for a logical ``AND``, you have to make multiple calls to the
              same operation.
 
@@ -1428,7 +1431,7 @@ partition
 
 With one or multiple callable, partition the items into 2 subgroups of items.
 
-.. warning:: The `callbacks` parameter is variadic and they are evaluated as a logical ``OR``.
+.. warning:: The `callbacks` parameter is variadic and will be evaluated as a logical ``OR``.
              If you're looking for a logical ``AND``, you have to make multiple calls to the
              same operation.
 
@@ -1752,7 +1755,7 @@ since
 
 Skip items until the callback is met.
 
-.. warning:: The `callbacks` parameter is variadic and they are evaluated as a logical ``OR``.
+.. warning:: The `callbacks` parameter is variadic and will be evaluated as a logical ``OR``.
              If you're looking for a logical ``AND``, you have to make multiple calls to the
              same operation.
 
@@ -1902,7 +1905,7 @@ Iterate over the collection items while the provided callback(s) are satisfied.
 
 It stops iterating when the callback(s) are not met.
 
-.. warning:: The `callbacks` parameter is variadic and they are evaluated as a logical ``OR``.
+.. warning:: The `callbacks` parameter is variadic and will be evaluated as a logical ``OR``.
              If you're looking for a logical ``AND``, you have to make multiple calls to the
              same operation.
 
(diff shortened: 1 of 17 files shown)


tests/static-analysis/contains.php
```diff
@@ -0,0 +1,43 @@
...
+contains_checkMap(Collection::fromIterable(['foo' => 'bar', 'baz' => 'taz'])->contains('bar'));
+/** @psalm-suppress InvalidArgument -> Psalm is smart and knows that 'abc' cannot be in there */
+contains_checkMap(Collection::fromIterable(['foo' => 'bar', 'baz' => 'taz'])->contains('bar', 'abc'));
+
+// VALID failures below -> `current` can return `NULL` if the collection is empty
+
+/** @psalm-suppress InvalidArgument @phpstan-ignore-next-line */
+contains_checkBool(Collection::fromIterable([1, 2, 3])->contains(2)->current());
+/** @psalm-suppress InvalidArgument @phpstan-ignore-next-line */
+contains_checkBool(Collection::fromIterable(['foo' => 'bar', 'baz' => 'taz'])->contains('bar')->current());
```
AlexandruGG, line 43: these feel very awkward to me. `current()` returns `T|null`, with `null` being returned when the collection is empty. However, in the chaining `contains()->current()` this will never be `null` because `contains()` will always return a collection with a boolean

spec/loophp/collection/CollectionSpec.php
```diff
@@ -671,6 +671,10 @@ public function it_can_contains(): void
         $this::fromIterable(range('A', 'C'))
             ->contains('C', 'unknown', 'A')
             ->shouldIterateAs([true]);
+
+        $this::fromIterable(['a' => 'b', 'c' => 'd'])
+            ->contains('d')
+            ->shouldIterateAs(['c' => true]);
```
AlexandruGG, line 677: this is another unexpected behaviour I discovered from `contains`, it doesn't always return a list, but rather keeps the key of the searched element

tests/static-analysis/contains.php
AlexandruGG: I guess in practice when used like this the analysers will not complain:

```PHP
if (true === Collection::fromIterable([1, 2, 3])->contains(2)->current()) {
    // do stuff
}
```

but if the comparison is not explicit PHPStan at least will complain because the return type is `bool|null`:

```PHP
if (Collection::fromIterable([1, 2, 3])->contains(2)->current()) {
    // do stuff
}```

File: grumphp.yml [hunk 3]
```diff
@@ -39,6 +39,16 @@ parameters:
...
+  tasks.psalm.ignore_patterns:
+      - "/.github/"
+      - "/.idea/"
+      - "/build/"
+      - "/benchmarks/"
+      - "/node_modules/"
+      - "/resource/"
+      - "/spec/"
+      - "/var/"
+      - "/vendor/"
```
AlexandruGG, line 51: I realised in CI Psalm wasn't running in `tests` where the static analysis checks are, so removing that ignore pattern here. I was still running them on local 😄

src/Collection.php
```diff
@@ -553,10 +553,6 @@ public function map(callable ...$callbacks): CollectionInterface
 
     public function match(callable $callback, ?callable $matcher = null): CollectionInterface
     {
-        // @todo: Rename this in next major version.
-        // We cannot use Match::class because PHP 8 has
-        // a new "match" function and we cannot use the same name.
-        // @See https://github.com/loophp/collection/issues/56
```
AlexandruGG, line 559: I don't think this `@todo` was needed here anymore, right? This method works in php 8

drupol: Indeed, it's a leftover.

tests/static-analysis/contains.php
AlexandruGG: @drupol I'm not sure I can convince you to make `contains` just return `bool` 😛. But it would be so much easier 😅

drupol (dismissed):
All good here, thanks !

AlexandruGG:
@drupol let me know if this helps you understand how I approach these static analysis and if it's helpful towards adding one for `when`. The PR is ready on my side, I'm going to add some more for commonly-used methods but on future PRs

docs/pages/api.rst
```diff
@@ -448,17 +450,24 @@ Signature: ``Collection::compact(...$values);``
 contains
 ~~~~~~~~
 
-Check if the collection contains one or more values.
+Check if the collection contains one or more values. Note that if multiple values are passed the operation
```
drupol, line 453: Maybe we could copy paste the warning box that explain this? For example, see the DropWhile operation: https://loophp-collection.readthedocs.io/en/latest/pages/api.html#dropwhile

AlexandruGG: I thought I remembered a similar one! Sure can!

Is this valid across all operations btw? Variadic parameters acting as `OR`

drupol: Thanks !

And yeah, I'm trying to make all the variadic as `OR` indeed.

tests/static-analysis/contains.php
drupol: It was like that in the early versions of the library then I changed. Maybe we should discuss about that in a dedicate PR ?

docs/pages/api.rst
AlexandruGG: added now. Had to make a tweak to specify that in this case you cannot use it as a logical AND by calling in succession because the collection will have just a `bool` after the first call

drupol:
> @drupol let me know if this helps you understand how I approach these static analysis and if it's helpful towards adding one for `when`. The PR is ready on my side, I'm going to add some more for commonly-used methods but on future PRs

Yeah go ahead, going to merge this right now.

drupol:
Thanks again @AlexandruGG !!!!

AlexandruGG:
@drupol btw the part with the variadic parameter made me think of writing a section in the documentation about _Principles of working with Collection_, with things like:

- most of the time operations will return a new Collection
- many operations allow tweaking the behaviour by the user providing a custom callback
- how to make the Collection behave eagerly via the `squash` operation
- how to test methods that use Collection (might not be straightforward for many that are not used to lazy behaviour!)

I might have others come to mind as well 😄. What do you think?

drupol:
It would be a big plus indeed! 

The more documentation we have, the better it is. There are indeed many behaviors that could be documented.

drupol:
> @drupol btw the part with the variadic parameter made me think of writing a section in the documentation about _Principles of working with Collection_, with things like:
> 
>     * most of the time operations will return a new Collection
> 
>     * mentioning the logical `OR` behaviour with variadic parameters
> 
>     * many operations allow tweaking the behaviour by the user providing a custom callback
> 
>     * how to make the Collection behave eagerly via the `squash` operation
> 
>     * how to test methods that use Collection (might not be straightforward for many that are not used to lazy behaviour!)
> 
> 
> I might have others come to mind as well smile. What do you think?
> Some of these are mentioned in the Readme or in the api documentation in different places but I'm thinking would be good to see them in one place

Yes! These are definitely the ADN of the library and it worth to have mentions in the README.

PR #108, opened 2021-06-13
Operation Interfaces return Collection Interface

AlexandruGG:
This PR corrects the return type on Operation Contracts (Interfaces), which return the Collection Contract (Interface), _not_ the Collection class implementation


AlexandruGG:
@drupol I'm surprised that analysers didn't complain about the mismatch between the `@return` annotation and the actual return type 🤷‍♂️

drupol (approved):
All good here!

drupol:
Thanks once again!

PR #109, opened 2021-06-15
Static Analysis Checks: Common Operations (II)

AlexandruGG:

- [x] filter
- [x] first
- [x] head
- [x] last

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -711,7 +711,7 @@ Signature: ``Collection::falsy();``
 filter
 ~~~~~~
 
-Filter collection items based on one or more callbacks.
+Filter collection items based on one or more callbacks. Multiple callbacks will be treated as a logical ``AND``.
 
 Interface: `Filterable`_
 
@@ -719,12 +719,14 @@ Signature: ``Collection::filter(callable ...$callbacks);``
 
 .. code-block:: php
 
-    $callback = static function($value): bool {
-        return 0 === $value % 3;
-    };
+    $divisibleBy3 = static fn($value): bool => 0 === $value % 3;
+    $divisibleBy6 = static fn($value): bool => 0 === $value % 6;
 
     $collection = Collection::fromIterable(range(1, 10))
-        ->filter($callback); // [3, 6, 9]
+        ->filter($divisibleBy3); // [3, 6, 9]
+
+    $collection = Collection::fromIterable(range(1, 10))
+        ->filter($divisibleBy3, $divisibleBy6); // [6]
 
 first
 ~~~~~
--- a/src/Operation/First.php
+++ b/src/Operation/First.php
@@ -24,6 +24,10 @@ final class First extends AbstractOperation
      */
     public function __invoke(): Closure
     {
-        return Head::of();
+        /** @var Closure(Iterator<TKey, T>): Generator<TKey, T> $head */
+        $head = Head::of();
+
+        // Point free style.
+        return $head;
     }
 }
--- a/src/Operation/First.php
+++ b/src/Operation/First.php
@@ -24,10 +24,6 @@ final class First extends AbstractOperation
      */
     public function __invoke(): Closure
     {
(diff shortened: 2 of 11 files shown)


docs/pages/api.rst
```diff
@@ -711,20 +711,22 @@ Signature: ``Collection::falsy();``
 filter
 ~~~~~~
 
-Filter collection items based on one or more callbacks.
+Filter collection items based on one or more callbacks. Multiple callbacks will be treated as a logical ``AND``.
```
AlexandruGG, line 714: I feel like it's a bit inconsistent with the others that `filter` works like this. I was expecting multiple callbacks to behave like logical `OR`

AlexandruGG: @drupol what do you think?

AlexandruGG:
Weird, looks like unit tests on windows on php 8 sometimes get stuck taking forever. Saw it happen once previously but I cannot cancel the job 🤷‍♂️. 10 min and still going

docs/pages/api.rst
drupol: Yep, this is one of the gotcha(s) I couldn't fix in a lazy way.

src/Operation/First.php
```diff
@@ -24,10 +24,6 @@ final class First extends AbstractOperation
      */
     public function __invoke(): Closure
     {
-        /** @var Closure(Iterator<TKey, T>): Generator<TKey, T> $head */
-        $head = Head::of();
-
-        // Point free style.
-        return $head;
+        return Head::of();
```
drupol, line 27: By removing this, you get 2 new minor errors in PSalm:

```
 ~/d/g/collection   feature/more-static-analysis-checks $…  ./vendor/bin/psalm --show-info=true src/Operation/First.php                                            
Scanning files...
Analyzing files...

░

INFO: MixedReturnTypeCoercion - src/Operation/First.php:23:16 - The declared return type 'Closure(Iterator<TKey:loophp\collection\Operation\First as mixed, T:loophp\collection\Operation\First as mixed>):Generator<TKey:loophp\collection\Operation\First as mixed, T:loophp\collection\Operation\First as mixed, mixed, mixed>' for loophp\collection\Operation\First::__invoke is more specific than the inferred return type 'Closure' (see https://psalm.dev/197)
     * @return Closure(Iterator<TKey, T>): Generator<TKey, T>


INFO: MixedReturnTypeCoercion - src/Operation/First.php:27:16 - The type 'Closure' is more general than the declared return type 'Closure(Iterator<TKey:loophp\collection\Operation\First as mixed, T:loophp\collection\Operation\First as mixed>):Generator<TKey:loophp\collection\Operation\First as mixed, T:loophp\collection\Operation\First as mixed, mixed, mixed>' for loophp\collection\Operation\First::__invoke (see https://psalm.dev/197)
        return Head::of();


------------------------------
No errors found!
------------------------------
2 other issues found.
------------------------------

Checks took 0.54 seconds and used 235.298MB of memory
Psalm was able to infer types for 100% of the codebase
 ~/d/g/collection   feature/more-static-analysis-checks $
```

drupol (commented):
Hello,
Most of the things are great, I just added a minor remark.
Let me know what you think.

docs/pages/api.rst
AlexandruGG: I see, no worries I've updated the docs and example 😁

src/Operation/First.php
AlexandruGG: oooh I didn't notice those, I guess because of the `info` level. I'll see if I can fix those, if not I'll revert

AlexandruGG: I see now why it happens, it's because `AbstractOperation::of` has the simple return type [Closure](https://github.com/loophp/collection/blob/master/src/Operation/AbstractOperation.php#L21). I think this makes sense as they are not all the same.

I don't think it should be a problem because it's just an info thing, but I've reverted this change to keep the same style as others

AlexandruGG:
thanks!

drupol:
Thanks you!!!

PR #111, opened 2021-06-16
Update Distinct operation

### LINKED ISSUE #110: Let users customize how to access a value in the Distinct operation.
The user should be able to use the `distinct` operation with a collection of object.

Currently, we use `===` in the Distinct operation to compare values, however, it happens sometimes that we want to customize
how to access to the values to compare.

Code example

<details>

```php
<?php

/**
 * For the full copyright and license information, please view
 * the LICENSE file that was distributed with this source code.
 */

declare(strict_types=1);

namespace App;

use loophp\collection\Collection;

include __DIR__ . '/vendor/autoload.php';

class User {
    private string $name;

    public function __construct(string $name)
    {
        $this->name = $name;
    }

    public function name(): string
    {
        return $this->name;
    }
};

$users = [
    new User('a'),
    new User('b'),
    new User('a'),
    new User('c'),
];

$collection = Collection::fromIterable($users)
    ->distinct();

print_r($collection->all());
```

</details>  

In this particular case, all the users will be returned and the Distinct function will not work.

## How to fix?

The idea is to let users customize how values should be accessed and/or compared.

### Proposal 1

Let users customize how to access to the value to compare.

```php
$collection = Collection::fromIterable($users)
    ->distinct(static fn(User $user): string => $user->name());
```

### Proposal 2

Let users customize how to access to the value to compare and how to compare them.

```php
$collection = Collection::fromIterable($users)
    ->distinct(static fn(User $left, User $right): bool => $left->name() === $right->name());
```

### Proposal 3

Mix the two previous proposals:

```php
$collection = Collection::fromIterable($users)
    ->distinct(
        static fn(User $user): string => $user->name(),
        static fn(string $left, string $right): bool => $left === $right
    ),
```

I tested both proposals and I'm unable to decide which way to go.

Therefore, this issue is to decide on which way to go.

drupol:
This PR

* [x] Update the Distinct operation
* [x] Has an updated documentation 
* [x] Is backward compatible  
* [x] Has updated tests
* [x] Has static tests

Fixes #110

### DIFF
--- a/docs/pages/code/operations/distinct.php
+++ b/docs/pages/code/operations/distinct.php
@@ -14,11 +14,11 @@ use loophp\collection\Collection;
 
 include __DIR__ . '/../../../../vendor/autoload.php';
 
-// Example 1 -> Using the default callback, with scalar values
+// Example 1 -> Using the default callbacks, with scalar values
 $collection = Collection::fromIterable(['a', 'b', 'a', 'c'])
     ->distinct(); // [0 => 'a', 1 => 'b', 3 => 'c']
 
-// Example 2 -> Using one custom callback, with object values
+// Example 2 -> Using a custom comparator callback, with object values
 final class User
 {
     private string $name;
@@ -46,7 +46,36 @@ $collection = Collection::fromIterable($users)
         static fn (User $left): Closure => static fn (User $right): bool => $left->name() === $right->name()
     ); // [0 => User<foo>, 1 => User<bar>, 3 => User<a>]
 
-// Example 3 -> Using two custom callbacks, with object values
+// Example 3 -> Using a custom accessor callback, with object values
+final class Person
+{
+    private string $name;
+
+    public function __construct(string $name)
+    {
+        $this->name = $name;
+    }
+
+    public function name(): string
+    {
+        return $this->name;
+    }
+}
+
+$users = [
+    new Person('foo'),
+    new Person('bar'),
+    new Person('foo'),
+    new Person('a'),
+];
+
+$collection = Collection::fromIterable($users)
+    ->distinct(
+        null,
+        static fn (Person $person): string => $person->name()
+    ); // [0 => Person<foo>, 1 => Person<bar>, 3 => Person<a>]
+
+// Example 4 -> Using both accessor and comparator callbacks, with object values
 final class Cat
 {
     private string $name;
--- a/docs/pages/code/operations/distinct.php
+++ b/docs/pages/code/operations/distinct.php
@@ -1,0 +1,76 @@
+<?php
+
+/**
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace App;
+
+use Closure;
+use loophp\collection\Collection;
+
+include __DIR__ . '/../../../../vendor/autoload.php';
+
+// Example 1 -> Using the default callback, with scalar values
+$collection = Collection::fromIterable(['a', 'b', 'a', 'c'])
+    ->distinct(); // [0 => 'a', 1 => 'b', 3 => 'c']
+
+// Example 2 -> Using one custom callback, with object values
+final class User
+{
+    private string $name;
+
+    public function __construct(string $name)
+    {
+        $this->name = $name;
+    }
+
+    public function name(): string
+    {
+        return $this->name;
+    }
+}
+
+$users = [
+    new User('foo'),
+    new User('bar'),
+    new User('foo'),
+    new User('a'),
+];
+
+$collection = Collection::fromIterable($users)
+    ->distinct(
+        static fn (User $left): Closure => static fn (User $right): bool => $left->name() === $right->name()
+    ); // [0 => User<foo>, 1 => User<bar>, 3 => User<a>]
+
+// Example 3 -> Using two custom callbacks, with object values
+final class Cat
+{
+    private string $name;
+
+    public function __construct(string $name)
+    {
+        $this->name = $name;
+    }
+
+    public function name(): string
+    {
+        return $this->name;
+    }
+}
+
+$users = [
+    new Cat('izumi'),
+    new Cat('nakano'),
+    new Cat('booba'),
+    new Cat('booba'),
+];
+
+$collection = Collection::fromIterable($users)
+    ->distinct(
+        static fn (string $left): Closure => static fn (string $right): bool => $left === $right,
+        static fn (Cat $cat) => $cat->name()
+    ); // [0 => Cat<izumi>, 1 => Cat<nakano>, 2 => Cat<booba>]
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -264,9 +264,30 @@ final class Collection implements CollectionInterface
         return new self(DiffKeys::of()(...$values), $this->getIterator());
     }
 
-    public function distinct(): CollectionInterface
+    public function distinct(?callable $comparatorCallback = null, ?callable $accessorCallback = null): CollectionInterface
     {
-        return new self(Distinct::of(), $this->getIterator());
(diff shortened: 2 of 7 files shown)


docs/pages/code/operations/distinct.php
```diff
@@ -0,0 +1,74 @@
...
+    new Cat('nakano'),
+    new Cat('booba'),
+    new Cat('booba'),
+];
+
+$collection = Collection::fromIterable($users)
+    ->distinct(
+        static fn (string $left): Closure => static fn (string $right): bool => $left === $right,
+        static fn (Cat $cat) => $cat->name()
+    ); // [0 => Cat<izumi>, 1 => Cat<nakano>, 2 => User<booba>]
```
AlexandruGG, line 74:
```suggestion
    ); // [0 => Cat<izumi>, 1 => Cat<nakano>, 2 => Cat<booba>]
```

src/Collection.php
```diff
@@ -263,9 +263,30 @@ public function diffKeys(...$values): CollectionInterface
...
+            /**
+             * @param mixed $left
+             *
+             * @return Closure(mixed): bool
+             */
+            static fn ($left): Closure =>
+                /**
+                 * @param mixed $right
+                 */
+                static fn ($right): bool => $left === $right;
```
AlexandruGG, line 287: would it not be cleaner to move these inside the `Distinct` operation?

drupol: We could but I don't think I ever used nullable parameter in the operations (I might be wrong), this is why I preprocess them before calling the Distinct operation.

AlexandruGG:
here's how to fix the phpstan issue: https://phpstan.org/r/a5fbb74a-9e50-401a-b11e-05d96cc477db

docs/pages/code/operations/distinct.php
```diff
@@ -0,0 +1,74 @@
...
+
+$users = [
+    new User('foo'),
+    new User('bar'),
+    new User('foo'),
+    new User('a'),
+];
+
+$collection = Collection::fromIterable($users)
+    ->distinct(static fn (User $user) => $user->name()); // [0 => User<foo>, 1 => User<bar>, 3 => User<a>]
```
AlexandruGG, line 45: how can the accessor callback be used as the first argument? The first argument is the comparator callback.

src/Collection.php
AlexandruGG: but you already made it nullable here https://github.com/loophp/collection/pull/111/files?diff=split&w=1#diff-a8a263aeb0e5677f2d5480899096e5077a07ba66c37c6d45378d39f745b3f219R26

drupol: Sure, in the collection, but not the operation itself.

AlexandruGG: oops, you're right, I was looking at the wrong thing. Ok makes sense then let's keep it here then

docs/pages/code/operations/distinct.php
drupol: Fixed, thanks !

docs/pages/code/operations/distinct.php
```diff
@@ -0,0 +1,76 @@
...
+    new User('foo'),
+    new User('bar'),
+    new User('foo'),
+    new User('a'),
+];
+
+$collection = Collection::fromIterable($users)
+    ->distinct(
+        static fn (User $left): Closure => static fn (User $right): bool => $left->name() === $right->name()
+    ); // [0 => User<foo>, 1 => User<bar>, 3 => User<a>]
```
AlexandruGG, line 47: can you add another example which just uses the accessor callback? I assume you'd do:

`->distinct(null, static fn (User $user) => $user->name())`

AlexandruGG: also would be good to have all cases in the spec:

- no callback
- only comparator callback
- only accessor callback
- both comparator and accessor callbacks

drupol: Ok ! I will continue to work on this in the evening.

drupol: - [x] no callback (it was there already)
- [x] only comparator callback (it was there already)
- [x] only accessor callback (just added in https://github.com/loophp/collection/pull/111/commits/875e2ae28c97dc3fc14a5b989436041266e6d250)
- [x] both comparator and accessor callbacks (it was there already)

drupol:
Do you have an idea on how I could fix the annotations here @AlexandruGG ?

AlexandruGG:
> Do you have an idea on how I could fix the annotations here @AlexandruGG ?

Hey sorry I only saw this now. If everything else is done leave this to me I can have a look and try to get it ready tomorrow

drupol:
Sure, no hurry!

src/Contract/Operation/Distinctable.php
```diff
@@ -19,8 +19,10 @@
 interface Distinctable
 {
     /**
-     * @param null|callable(mixed): (Closure(mixed): bool) $comparatorCallback
-     * @param null|callable(T, TKey): mixed $accessorCallback
+     * @template U
+     *
+     * @param null|callable(U): (Closure(U): bool) $comparatorCallback
+     * @param null|callable(T, TKey): U $accessorCallback
```
AlexandruGG: note a very important element: the type returned by the accessor callback needs to match the type of the parameters in the closures from the comparator callback, otherwise a type error will happen at runtime

src/Operation/Distinct.php
```diff
@@ -22,47 +22,60 @@
...
+                 * @param callable(T, TKey): U $accessorCallback
                  *
                  * @return Closure(Iterator<TKey, T>): Generator<TKey, T>
                  */
                 static function (callable $accessorCallback) use ($comparatorCallback): Closure {
+                    /**
+                     * @param callable(T, TKey): U $accessorCallback
+                     *
+                     * @return Closure(callable(U): Closure(U): bool): Closure(list<array{0: TKey, 1: T}>, array{0: TKey, 1: T}): list<array{0: TKey, 1: T}>
+                     */
```
AlexandruGG, line 48: this as well as the return type [here](https://github.com/loophp/collection/pull/111/commits/c0fca86fb35e00fac05d8e9288b8f6b301e1bafc#diff-3b63570fb11bc48cd14c56240059ece0e699ed0f466f9ea7ab7c10ba9c3616abR54) were the missing pieces that were causing the Psalm static analysis errors

AlexandruGG (approved):

AlexandruGG:
@drupol this is sorted now, have a look at my last commit https://github.com/loophp/collection/pull/111/commits/c0fca86fb35e00fac05d8e9288b8f6b301e1bafc to see the changes done.

Maybe we can have a chat sometime so I can explain how I approach the static analysis checks to ensure that we get the types right and that we protect against incorrect usage (otherwise failures will only be detected by tests / at runtime!).

drupol:
This, is my current state of mind:

Image

Going to read this carefully and learn! :)

AlexandruGG:
> This, is my current state of mind:
> 
> Image
> 
> Going to read this carefully and learn! :)

😄  that was legitimately my feeling when I first found your library and looked into the code! I had never worked with so many closures and currying in PHP!

drupol:
> smile that was legitimately my feeling when I first found your library and looked into the code! I had never worked with so many closures and currying in PHP!

Haha thanks mate! Glad you like it and that you learned something! That's the spirit :)

src/Operation/Distinct.php
drupol: I got it now, my bad!

drupol:
> Maybe we can have a chat sometime so I can explain how I approach the static analysis checks to ensure that we get the types right and that we protect against incorrect usage (otherwise failures will only be detected by tests / at runtime!). It can be tricky with all the closures sometimes though sweat_smile

Great idea, anytime !!

PR #112, opened 2021-06-19
Introduce MapN | Deprecate Map With Multiple Callbacks | Static Analysis Checks

AlexandruGG:
This PR introduces `MapN` based on the discussions in https://github.com/loophp/collection/pull/77, and deprecates `Map` usage with multiple callbacks.

* [x] `MapN` code (same as `Map` but with more `mixed` type hints)
* [x] Deprecate `Map` with multiple callbacks
* [x] `Map` stays backwards compatible
* [x] Tests
* [x] Static analysis checks
* [x] Documentation

### DIFF
--- a/spec/loophp/collection/CollectionSpec.php
+++ b/spec/loophp/collection/CollectionSpec.php
@@ -1866,9 +1866,9 @@ class CollectionSpec extends ObjectBehavior
             ->shouldIterateAs(['1', '4', '9']);
 
         $this::fromIterable(range(1, 3))
-            ->map(static fn (int $item): int => $item ** 2)
-            ->map(static fn (int $item): string => (string) $item)
-            ->map(static fn (string $item): string => $item . 'bar')
+            ->mapN(static fn (int $item): int => $item ** 2)
+            ->mapN(static fn (int $item): string => (string) $item)
+            ->mapN(static fn (string $item): string => $item . 'bar')
             ->shouldIterateAs(['1bar', '4bar', '9bar']);
     }
 
--- a/spec/loophp/collection/CollectionSpec.php
+++ b/spec/loophp/collection/CollectionSpec.php
@@ -21,7 +21,6 @@ use loophp\collection\Collection;
 use loophp\collection\Contract\Operation;
 use loophp\collection\Operation\AbstractOperation;
 use OutOfBoundsException;
-use PhpSpec\Exception\Example\ErrorException;
 use PhpSpec\Exception\Example\FailureException;
 use PhpSpec\Exception\Example\MatcherException;
 use PhpSpec\ObjectBehavior;
--- a/spec/loophp/collection/CollectionSpec.php
+++ b/spec/loophp/collection/CollectionSpec.php
@@ -22,11 +22,13 @@ use loophp\collection\Collection;
 use loophp\collection\Contract\Operation;
 use loophp\collection\Operation\AbstractOperation;
 use OutOfBoundsException;
+use PhpSpec\Exception\Example\ErrorException;
 use PhpSpec\Exception\Example\FailureException;
 use PhpSpec\Exception\Example\MatcherException;
 use PhpSpec\ObjectBehavior;
 use stdClass;
 use function gettype;
+use const E_USER_DEPRECATED;
 use const INF;
 use const PHP_EOL;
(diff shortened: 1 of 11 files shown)


src/Contract/Operation/Mapable.php
```diff
@@ -15,12 +15,15 @@
...
- * @template V
  */
 interface Mapable
 {
     /**
      * Apply one or more callbacks to a collection and use the return value.
+     * Usage with multiple callbacks is deprecated and will be removed in a future major version.
+     * Use mapN instead for multiple callbacks.
+     *
+     * @template V
```
AlexandruGG, line 26: the template for the return type needs to be on the method, not on the interface, because it comes from the callback passed to the method. That's why there was [this error](https://github.com/loophp/collection/pull/112/files?diff=split&w=1#diff-995edee38ad4f8387e58ebd52c31bcc04c56cc2448d331b1cf5e0b35c57b9efaL9-L12) reported before

AlexandruGG:
@drupol hey, let me know if this looks good to you and I will update the documentation as well

src/Contract/Operation/Mapable.php
drupol: Yep exactly that!

spec/loophp/collection/CollectionSpec.php
```diff
@@ -1822,28 +1825,51 @@ public function it_can_map(): void
...
+
+        $square = static fn (int $a): int => $a ** 2;
+        $toString = static fn (int $a): string => (string) $a;
+
+        $this::fromIterable(range(1, 3))
+            ->mapN($square, $toString)
+            ->shouldIterateAs(['1', '4', '9']);
+
+        $this::fromIterable(range(1, 3))
+            ->map(static fn (int $item): int => $item ** 2)
```
drupol, line 1869: You're using `::map()` and we are in `it_can_mapN`

drupol (changes requested):
Hi Alex,

Thanks for this PR !
I was about to give the approval but then I had an idea.

Instead of throwing the exception in the `Operation class Map`, why not throwing it from `Collection::map()` ?
Wouldn't it be better? WDYT?

spec/loophp/collection/CollectionSpec.php
AlexandruGG: good spot, copy-paste mistake 😅 . Will fix

AlexandruGG:
> Hi Alex,
> 
> Thanks for this PR !
> I was about to give the approval but then I had an idea.
> 
> Instead of throwing the exception in the `Operation class Map`, why not throwing it from `Collection::map()` ?
> Wouldn't it be better? WDYT?

Hey, thanks for reviewing!

I considered that, however it's not sufficient because the library provides the Operations to be used separately. So someone that will use the operation directly [like this](https://github.com/loophp/collection/blame/master/docs/index.rst#L116-L121) also needs to know about the deprecation if used with multiple callbacks. That's why I also had to update [this usage](https://github.com/loophp/collection/pull/112/files#diff-654f3fe07b79a5130f70435648f62717f2ede0b49e5b466b9bb5a6c5f55c8a2eR53) to `MapN`

drupol:
> > Hi Alex,
> > Thanks for this PR !
> > I was about to give the approval but then I had an idea.
> > Instead of throwing the exception in the `Operation class Map`, why not throwing it from `Collection::map()` ?
> > Wouldn't it be better? WDYT?
> 
> Hey, thanks for reviewing!
> 
> I considered that, however it's not sufficient because the library provides the Operations to be used separately. So someone that will use the operation directly [like this](https://github.com/loophp/collection/blame/master/docs/index.rst#L116-L121) also needs to know about the deprecation if used with multiple callbacks. That's why I also had to update [this usage](https://github.com/loophp/collection/pull/112/files#diff-654f3fe07b79a5130f70435648f62717f2ede0b49e5b466b9bb5a6c5f55c8a2eR53) to `MapN`

You're perfectly right, that was a bad idea :)

Can't wait to be in version 5 to remove the ugly `count()` in `map`!

AlexandruGG:
@drupol hey, added documentation now and addressed comments, this should be ready 😁

drupol (approved):
Nothing to add here, all good :)

PR #114, opened 2021-06-21
Static Analysis Checks: Common Operations (III)

AlexandruGG:

- [x] every
- [x] get
- [x] key
- [x] keys
- [x] normalize
- [x] squash

Follows #109.

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -439,6 +439,10 @@ final class Collection implements CollectionInterface
         );
     }
 
+    /**
+     * @param TKey $key
+     * @param T|null $default
+     */
     public function get($key, $default = null): CollectionInterface
     {
         return new self(Get::of()($key)($default), $this->getIterator());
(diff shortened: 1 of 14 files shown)


src/Collection.php
```diff
@@ -439,6 +439,10 @@ public static function fromString(string $string, string $delimiter = ''): self
         );
     }
 
+    /**
+     * @param TKey $key
+     * @param T|null $default
+     */
     public function get($key, $default = null): CollectionInterface
```
AlexandruGG, line 446: the `@param T|null $default` annotation is needed here because otherwise Psalm interprets $default as type `T` and `T` as `null` because of the default null value.

Because of that it thinks that `Collection::fromIterable([1, 2, 3])->get(1)` is always `Collection<int, null>` instead of `Collection<int, int|null>`

drupol: Fair enough, good for me.

drupol:
All good :) Thanks !

PR #115, opened 2021-06-24
feat: Add Matching operation and implements Selectable from Doctrine.

drupol:
This PR

* [x] Implements `Collection::matching(Criteria $criteria)` operation
* [x] Has tests
* [x] Update documentation 
* [x] Fix static analysis <- Do we need them?

To investigate:

* [x] Should we detach each row from Doctrine? (https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/batch-processing.html#iterating-large-results-for-data-processing)

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -1334,6 +1334,21 @@ Signature: ``Collection::match(callable $callback, ?callable $matcher = null);``
 .. literalinclude:: code/operations/match.php
   :language: php
 
+matching
+~~~~~~~~
+
+Collection implements the ``Selectable`` interface provided by `Doctrine Collections`_.
+
+That means it let you use the Criteria API provided by Doctrine,
+in a complete lazy way.
+
+Interface: `Matchingable`_
+
+Signature: ``Collection::matching(Criteria $criteria);``
+
+.. literalinclude:: code/operations/matching.php
+  :language: php
+
 merge
 ~~~~~
 
@@ -2304,6 +2319,7 @@ Signature: ``Collection::zip(iterable ...$iterables);``
         ->limit(100)
         ->unwrap(); // [0, 1, 2, 3 ... 196, 197, 198, 199]
 
+.. _Doctrine Collections: https://github.com/doctrine/collections
 .. _Allable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Allable.php
 .. _Appendable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Appendable.php
 .. _Applyable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Applyable.php
@@ -2320,6 +2336,7 @@ Signature: ``Collection::zip(iterable ...$iterables);``
 .. _Containsable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Containsable.php
 .. _Countable: https://www.php.net/manual/en/class.countable.php
 .. _Currentable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Currentable.php
+.. _Criteria: https://www.doctrine-project.org/projects/doctrine-collections/en/1.6/index.html#matching
 .. _Cycleable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Cycleable.php
 .. _Diffable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Diffable.php
 .. _Diffkeysable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Diffkeysable.php
@@ -2362,6 +2379,7 @@ Signature: ``Collection::zip(iterable ...$iterables);``
 .. _Mapable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Mapable.php
 .. _MapNable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/MapNable.php
 .. _Matchable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Matchable.php
+.. _Matchingable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Matchingable.php
 .. _Mergeable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Mergeable.php
 .. _Normalizeable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Normalizeable.php
 .. _Nthable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Nthable.php
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -1337,10 +1337,8 @@ Signature: ``Collection::match(callable $callback, ?callable $matcher = null);``
 matching
 ~~~~~~~~
 
-Collection implements the ``Selectable`` interface provided by `Doctrine Collections`_.
-
-That means it lets you use the Criteria API provided by Doctrine,
-in a completely lazy way.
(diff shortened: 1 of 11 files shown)


AlexandruGG:
tag me when this is ready, I'll want to review :)

docs/pages/api.rst
```diff
@@ -1334,6 +1334,21 @@ Signature: ``Collection::match(callable $callback, ?callable $matcher = null);``
 .. literalinclude:: code/operations/match.php
   :language: php
 
+matching
+~~~~~~~~
+
+Collection implements the ``Selectable`` interface provided by `Doctrine Collections`_.
+
+That means it let you use the Criteria API provided by Doctrine,
+in a complete lazy way.
```
AlexandruGG, line 1343:
```suggestion
That means it lets you use the Criteria API provided by Doctrine,
in a completely lazy way.
```

src/Contract/Operation/Matchingable.php
```diff
@@ -0,0 +1,28 @@
...
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace loophp\collection\Contract\Operation;
+
+use Doctrine\Common\Collections\Criteria;
+use Doctrine\Common\Collections\Selectable as DoctrineCollectionsSelectable;
```
AlexandruGG, line 13:
```suggestion
use Doctrine\Common\Collections\Criteria;
use Doctrine\Common\Collections\Selectable;
```

I don't see why the alias is needed if we don't have an alias for the other import

src/Contract/Operation/Matchingable.php
```diff
@@ -0,0 +1,28 @@
...
+use loophp\collection\Contract\Collection;
+
+/**
+ * @template TKey
+ * @template T
+ *
+ * @template-extends DoctrineCollectionsSelectable<TKey, T>
```
AlexandruGG, line 20:
```suggestion
 * @extends DoctrineCollectionsSelectable<TKey, T>
```

have you tried it like this? Could this be causing the errors in the baseline?

phpstan-baseline.neon
```diff
@@ -1,5 +1,25 @@
...
+
+		-
+			message: "#^Return type \\(loophp\\\\collection\\\\Contract\\\\Collection\\) of method loophp\\\\collection\\\\Collection\\:\\:matching\\(\\) should be covariant with return type \\(loophp\\\\collection\\\\Contract\\\\Collection\\<mixed, mixed\\>\\) of method loophp\\\\collection\\\\Contract\\\\Operation\\\\Matchingable\\<TKey,T\\>\\:\\:matching\\(\\)$#"
+			count: 2
+			path: src/Collection.php
+
+		-
+			message: "#^Return type \\(loophp\\\\collection\\\\Contract\\\\Collection\\<TKey, T\\>\\) of method loophp\\\\collection\\\\Contract\\\\Operation\\\\Matchingable\\:\\:matching\\(\\) should be covariant with return type \\(Doctrine\\\\Common\\\\Collections\\\\Collection\\<mixed, mixed\\>\\) of method Doctrine\\\\Common\\\\Collections\\\\Selectable\\<TKey,T\\>\\:\\:matching\\(\\)$#"
+			count: 1
+			path: src/Contract/Operation/Matchingable.php
```
AlexandruGG, line 21: I don't think we should add these to the baseline, something seems to be wrong with the template annotations. Also, a static analysis check in `tests/` would be good - should be fairly simple because the type of the items in the collection doesn't change

AlexandruGG: let me know if you need me to take a stab at these 😅

AlexandruGG:
>  Should we detach each row from Doctrine? (https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/batch-processing.html#iterating-large-results-for-data-processing)

Regarding this, I think it could be implemented in a follow-up PR with a callback provided by the user. I don't think it should be done "by default", and also we don't have access to the entity manager, right? The user would have to provide that

src/Contract/Operation/Matchingable.php
drupol: Yep, it doesn't change anything.

phpstan-baseline.neon
drupol: I don't know if we can fix so easily

drupol: Yeah if you could have a look, I'm unable to get rid of those messages.

AlexandruGG: ok I will look into it today and update you

src/Contract/Operation/Matchingable.php
```diff
@@ -0,0 +1,28 @@
...
+use Doctrine\Common\Collections\Selectable;
+use loophp\collection\Contract\Collection;
+
+/**
+ * @template TKey
+ * @template T
+ *
+ * @extends Selectable<TKey, T>
+ */
+interface Matchingable extends Selectable
```
AlexandruGG, line 22:
```suggestion
interface Matchingable
```

@drupol hey, I've looked into the static analysis issues, and here is the problem: we cannot make Matchingable to extend Selectable, because there is no relationship between loophp Collection and Doctrine Collection, so this would break [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle). This is why we're seeing these errors related to covariance reported by PHPStan: https://github.com/loophp/collection/pull/115/files#diff-995edee38ad4f8387e58ebd52c31bcc04c56cc2448d331b1cf5e0b35c57b9efaR9-R19

All errors (both Psalm and PHPStan) disappear if we remove this extending, which I think is what we should do

drupol: Oh right. I'm mitigated about this.

1. I'm for it because [the return type of Selectable](https://github.com/doctrine/collections/blob/1.6.x/lib/Doctrine/Common/Collections/Selectable.php#L28) is `@psalm-return Collection<TKey,T>` and it doesn't correspond to anything.
2. I'm against it because `Selectable` is something that people are used to manipulate and it would ease the adoption in other places. If we don't do that, I don't see the point of implementing it.

AlexandruGG: unless loophp/collection implements the Doctrine Collection interface, I don't think we should extend `Selectable`. I think there's nothing wrong with having `Matchingable` by itself.

Is your concern related to this use case?

```PHP
function getResults(Selectable $selectable, Criteria $criteria): Doctrine\Collection
{
    // some other code
    return $selectable->matching($criteria);
}
```

For users that are using `matching` like the above, regardless of whether we extend `Selectable` the issue will still be regarding the return type of `getResults`. In their current usage the selectable will return **Doctrine Collection**, however if they try to just pass in **loophp/Collection** into this function, the return type changes to loophp Collection. This might be unexpected and will actually lead to them having to change the return type to `loophp\Collection | Doctrine\Collection` anyway.

If we _don't_ extend Selectable, then this fact is just made a bit more obvious, and they will have to update the function accordingly:

```PHP
/**
* @param Selectable | Matchingable $selectable
*
* @return Doctrine\Collection | loophp\Collection
*/
function getResults(Selectable $selectable, Criteria $criteria)
{
    // some other code
    return $selectable->matching($criteria);
}
```

drupol: Valid point. I will update accordingly in an hour.

drupol: Fixed. Thanks!

AlexandruGG (approved):
Nice work on this 💯

PR #117, opened 2021-06-27
Add FlatMap Operation

AlexandruGG:
This PR adds the `FlatMap` Operation.

* [x] has tests 
* [x] has static analysis check
* [x] has documentation
* [x] clarify key types and `normalize` usage

Bonus, static analysis checks:
- [x] flatten
- [x] unwrap

### DIFF
--- a/spec/loophp/collection/CollectionSpec.php
+++ b/spec/loophp/collection/CollectionSpec.php
@@ -1142,20 +1142,48 @@ class CollectionSpec extends ObjectBehavior
     public function it_can_flatMap(): void
     {
         $this::fromIterable([1, 2, 3])
+            ->flatMap(static fn (int $item, int $key): array => [$key => $item * $item])
+            ->shouldIterateAs([1, 4, 9]);
+
+        $gen = static function (): Generator {
+            yield 0 => 1;
+
+            yield 0 => 4;
+
+            yield 0 => 9;
+        };
+
+        $this::fromIterable([1, 2, 3])
+            ->flatMap(static fn (int $item): array => [$item * $item])
+            ->shouldIterateAs($gen());
+
+        $this::fromIterable([1, 2, 3])
             ->flatMap(static fn (int $item): array => [$item + $item, $item * $item])
+            ->normalize()
             ->shouldIterateAs([2, 1, 4, 4, 6, 9]);
 
         $this::fromIterable([1, 2, 3])
             ->flatMap(static fn (int $item): array => [[$item + $item], [$item * $item]])
+            ->normalize()
             ->shouldIterateAs([[2], [1], [4], [4], [6], [9]]);
 
         $this::fromIterable(['foo' => 'f', 'bar' => 'b'])
             ->flatMap(static fn (string $item, string $key): array => [$item => $key])
-            ->shouldIterateAs([0 => 'foo', 1 => 'bar']);
+            ->shouldIterateAs(['f' => 'foo', 'b' => 'bar']);
+
+        $gen = static function (): Generator {
+            yield 0 => ['f' => 'foo'];
+
+            yield 1 => ['FOO' => 'F'];
+
+            yield 0 => ['b' => 'bar'];
+
+            yield 1 => ['BAR' => 'B'];
+        };
 
         $this::fromIterable(['foo' => 'f', 'bar' => 'b'])
             ->flatMap(static fn (string $item, string $key): array => [[$item => $key], [mb_strtoupper($key) => mb_strtoupper($item)]])
-            ->shouldIterateAs([['f' => 'foo'], ['FOO' => 'F'], ['b' => 'bar'], ['BAR' => 'B']]);
+            ->shouldIterateAs($gen());
     }
 
     public function it_can_flatten(): void
--- a/spec/loophp/collection/CollectionSpec.php
+++ b/spec/loophp/collection/CollectionSpec.php
@@ -1139,6 +1139,25 @@ class CollectionSpec extends ObjectBehavior
             ->shouldIterateAs([true]);
     }
 
+    public function it_can_flatMap(): void
+    {
+        $this::fromIterable([1, 2, 3])
+            ->flatMap(static fn (int $item): array => [$item + $item, $item * $item])
+            ->shouldIterateAs([2, 1, 4, 4, 6, 9]);
+
+        $this::fromIterable([1, 2, 3])
+            ->flatMap(static fn (int $item): array => [[$item + $item], [$item * $item]])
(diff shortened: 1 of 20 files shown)


spec/loophp/collection/CollectionSpec.php
```diff
@@ -1139,6 +1139,25 @@ static function ($value) {
...
+            ->flatMap(static fn (int $item): array => [$item + $item, $item * $item])
+            ->shouldIterateAs([2, 1, 4, 4, 6, 9]);
+
+        $this::fromIterable([1, 2, 3])
+            ->flatMap(static fn (int $item): array => [[$item + $item], [$item * $item]])
+            ->shouldIterateAs([[2], [1], [4], [4], [6], [9]]);
+
+        $this::fromIterable(['foo' => 'f', 'bar' => 'b'])
+            ->flatMap(static fn (string $item, string $key): array => [$item => $key])
+            ->shouldIterateAs([0 => 'foo', 1 => 'bar']);
```
AlexandruGG, line 1154: @drupol I'd like to clarify this case because I think this initial implementation is a bit faulty.

Here I would expect to get `['f' => 'foo', 'b' => 'bar']`, however, because of the usage of `Normalize`, the string keys get replaced with integer ones. I don't think `flatMap` should change the type of keys, it's not something that this operation typically does.

I reckon we would need to use `Pack` and `Unpack` in such a way that the keys are maintained, right? I tried a couple of things but could not get it working properly as the map callback needs to be changed. What do you think?

drupol: Hi Alex!

Yes, we can safely remove the `normalize` I have no objection against it.

AlexandruGG: the problem is just by removing it the operation doesn't work 😄 . Because the indexes are the same, so `Collection::fromIterable([1, 2, 3])->flatMap(static fn (int $item): array => [$item * $item])` will be `[0 => 9]` instead of `[1, 4, 9]`

drupol: Let me get back to you after family lunch and I'll check with my computer.

drupol: Ok here's my thoughts.

I think we should, at all cost try to keep the association `[$key, $value]`, even if this is absurd... but this is how PHP is meant to be used.
This library try to keep this association almost everywhere, despite the fact that this could be discussed.

Why? because the user is able to use methods `::pack()` or `::wrap()` before calling this operation.
Using such methods, will change the structure of the data passed in the subsequent calls, but it's the user who's responsible of that, so he needs to act accordingly.

Anyway, the solution here is very simple.

If you really want to deal with keys, then your example needs to be updated like this:

```php
Collection::fromIterable([1, 2, 3])
    ->flatMap(static fn (int $item, int $key): array => [$key => $item * $item]);
```

The `::map()` method is able to deal with keys and values, so you have to provide it.

Let me know what you think.

AlexandruGG: Hmm at first I was a bit confused by this but I now see what you mean. It's best to not include `normalize` in this operation because it restricts the way it can be used when it comes to string keys. And when needed, the user can call that separately. These two would then be equivalent:

```PHP
Collection::fromIterable([1, 2, 3])
    ->flatMap(static fn (int $item, int $key): array => [$key => $item * $item]);

Collection::fromIterable([1, 2, 3])
    ->flatMap(static fn (int $item, int $key): array => [$item * $item])
    ->normalize();
```

Also this will work without normalize:

```PHP
$this::fromIterable(['foo' => 'f', 'bar' => 'b'])
            ->flatMap(static fn (string $item, string $key): array => [$item => $key])
            ->shouldIterateAs(['f' => 'foo', 'b' => 'bar']);

Comment (drupol) [reply spec/loophp/collection/CollectionSpec.php:1154]: Exactly, you got my point.

File: src/Operation/FlatMap.php [hunk 2]
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return
+            /**
+             * @param callable(T, TKey, Iterator<TKey, T>): T $callback
+             */
+            static function (callable $callback): Closure {
+                /** @var Closure(Iterator<TKey, T>): Generator<TKey, T> $flatMap */
+                $flatMap = Pipe::of()(
+                    Map::of()($callback),
+                    Unwrap::of()
```
drupol, line 35: `Unwrap::of()` is basically a shortcut to `Flatten::of()(1)`, maybe we should use it to get an extremelly minor perf improvement? WDYT ?

AlexandruGG: hmm sure I can use that, I didn't think about the performance 😄

drupol: excellent :+1:

drupol:
I really like the fact that the map callback can return any type of iterable, this will still continue to work !

src/Contract/Operation/FlatMapable.php
```diff
@@ -0,0 +1,31 @@
...
+ * @template T
+ */
+interface FlatMapable
+{
+    /**
+     * Transform the collection through a callback and flatten one level.
+     *
+     * @template V
+     *
+     * @param callable(T, TKey, Iterator<TKey, T>): V $callback
```
drupol, line 26: Shouldn't this be:

```suggestion
     * @param callable(T, TKey, Iterator<TKey, T>): iterable<IterableKey, IterableValue> $callback
```

AlexandruGG:
> I really like the fact that the map callback can return any type of `iterable`, this will still continue to work !

Nice observation! I'll add some unit tests to the relevant operations to ensure that behaviour doesn't break by accident

src/Contract/Operation/FlatMapable.php
AlexandruGG: yeah the types are not yet right, I will get them right when I add the static analysis check :)

src/Contract/Operation/Flattenable.php
```diff
@@ -22,7 +22,7 @@ interface Flattenable
     /**
      * Flatten a collection of items into a simple flat collection.
      *
-     * @return Collection<TKey, T>
+     * @return Collection<mixed, mixed>
```
AlexandruGG, line 25: unfortunately for `flatten` and `unwrap` these cannot be more specifically typed because the final collection will vary wildly depending on the shape of the data it had in it before applying these operations

AlexandruGG:
@drupol I think this is ready but have a look at the types, tests, and docs and see if everything is as if you'd expect from this

drupol:
All good, maybe next time it would be better to separate a new feature from the small refactoring, I prefer when PRs are clearly scoped.

PR #121, opened 2021-07-05
Trim Static Analysis Baselines

AlexandruGG:
This PR removes the Psalm baseline and the PHPStan Docs baseline, which is not in use anymore because the `docs` folder is ignored.


src/Iterator/RandomIterator.php
```diff
@@ -79,9 +79,10 @@ public function key()
 
     public function next(): void
     {
-        unset($this->indexes[$this->key]);
-
-        $this->key = key($this->indexes);
+        if (null !== $this->key) {
+            unset($this->indexes[$this->key]);
+            $this->key = key($this->indexes);
+        }
```
AlexandruGG, line 85: @drupol let me know if this looks right to you. If the key is _null_ then the `unset` will not have any effect anyway

drupol: Nice catch :)

AlexandruGG: Psalm's catch 😄

AlexandruGG:
thanks!

drupol:
Thanks mate!

PR #122, opened 2021-07-05
Show Collection is Immutable & Operations Pure

AlexandruGG:
Collection claims to make use of pure functions and immutability; this PR introduces annotations so that this can be verified via static analysis.

Note: there is still scope to improve on this and use `pure-callable` instead of `callable` in places where a pure callable should be provided: https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-pure.

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -136,6 +136,8 @@ use const PHP_INT_MIN;
  * @template TKey
  * @template T
  *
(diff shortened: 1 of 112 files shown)


AlexandruGG:
@drupol I think you'll like this one 😉

drupol:
Thanks!

PR #124, opened 2021-07-06
Refactor `::partition()` and `::span()`

### LINKED ISSUE #123: `Partition` Operation - Awkward to use?
## Background

I am looking to replace usage of Doctrine Collections with this library in a project, and one of the issues I've come across is the usage of `partition`. Its return type is a bit confusing at first glance and replicating the functionality of the Doctrine Collections method does not seems straightforward for a new user.

## Doctrine Collection

The return type of the [Doctrine Collection Partition](https://github.com/doctrine/collections/blob/1.6.x/lib/Doctrine/Common/Collections/Collection.php#L263) method is very straightforward. It returns an array shape with _exactly two elements_ consisting of the same collection objects with the same  type for the keys and values. Because of this, usage with the PHP [list](https://www.php.net/manual/en/function.list.php) language construct is possible and very convenient:

```PHP
$data = [
    ['id' => 'ABC', 'name' => 'Bob'],
    ['id' => 'BAC', 'name' => 'John'],
    ['id' => 'ABC', 'name' => 'Mike'],
];

[$matches, $nonMatches] = (new ArrayCollection($data))
    ->partition(static fn (int $key, array $entity): bool => $entity['id'] === 'ABC');
```

## Loophp Collection

The return type of the [partition operation](https://github.com/loophp/collection/blob/master/src/Contract/Operation/Partitionable.php#L23) is a collection of lists of arrays of an array shape where the key and value of the elements is included (the nesting is real!).

The first issue is that this operation will always return exactly two elements, but this is not immediately visible from the return type and I'm not sure how this can be fixed because we cannot specify collection shape in the same way as array shape.

The other main issue is the combination of operations needed to achieve the same behaviour as Doctrine Collections: _I just want to be able to get the two separate collections with the values inside them, nothing more._ This is what I've come up with:

```PHP
$partitioned = Collection::fromIterable($data)
    ->partition(static fn(array $entity): bool => $entity['id'] === 'ABC');

$matches = $partitioned->first()->pluck('*.1')->unwrap();
$nonMatches = $partitioned->last()->pluck('*.1')->unwrap();
```

## Possible improvements

A) don't return the keys and the values separately. Rather, they should be together just like in the original collection, before the partition
B) can we reduce the nesting so the `unwrap()` is not necessary?
C) could we return an array shape so that the `list` construct can be used? I realise this is a significant difference, however otherwise we rely on the user "knowing" that this operation will only return two elements in the collection - something that's not currently captured in the type annotations

drupol:
This PR

* [x] Refactor `::partition()` operation
* [x] Refactor `::span()` operation 
* [x] Has updated tests
* [x] Has static analysis tests

Depends/Follow on #125 
Fixes #123

### DIFF
--- a/src/Contract/Operation/Spanable.php
+++ b/src/Contract/Operation/Spanable.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 
 namespace loophp\collection\Contract\Operation;
 
+use Iterator;
 use loophp\collection\Contract\Collection;
 
 /**
@@ -18,7 +19,9 @@ use loophp\collection\Contract\Collection;
 interface Spanable
 {
     /**
-     * @return Collection<TKey, T>
+     * @param callable(T, TKey, Iterator<TKey, T>): bool $callback
+     *
+     * @return Collection<int, Collection<TKey, T>>
      */
     public function span(callable $callback): Collection;
 }
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -1509,11 +1509,24 @@ Signature: ``Collection::pair();``
 partition
 ~~~~~~~~~
 
-With one or multiple callable, partition the items into 2 subgroups of items.
+With one or multiple callable, partition the items into two subgroups of items.
 
 .. warning:: The `callbacks` parameter is variadic and will be evaluated as a logical ``OR``.
              If you're looking for a logical ``AND``, you have to make multiple calls to the
              same operation.
+
+The raw `Partition` operation returns a generator yielding two iterators only.
+
+The first inner iterator is the result of a `filter` operation, it contains items
+who has met the provided callback.
+The second and last inner is the result of a `reject` operation, it contains items
+who has not met the provided callback.
+
+When the `partition` operation is used through the `Collection` object, the two
+resulting iterators will be converted and mapped into a `Collection` object.
+
+The first inner collection contains items who has met the provided callback.
+The second and last collection contains items who has not met the provided callback.
 
 Interface: `Partitionable`_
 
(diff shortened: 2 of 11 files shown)


AlexandruGG:
Really nice getting this up so quickly! I will focus on the other PRs first so we can get those merged

drupol:
I found a way to optimize the thing and remove the entries I was about to commit in `psalm-baseline.xml` !

drupol:
Sorry for the back'n'forth, I don't touch this anymore now!

AlexandruGG:
> Sorry for the back'n'forth, I don't touch this anymore now!

It's okay I hadn't reviewed yet 😄 . There is no change on `Span` here other than the docs, did you decide to keep it for a separate PR or does nothing need to be refactored?

drupol:
The raw span operation is properly done, I just had to map the iterators to collection in Collection.php and it's good !

AlexandruGG:
> The raw span operation is properly done, I just had to map the iterators to collection in Collection.php and it's good !

Sounds good, thanks for moving so fast on this. Leave this to me then I will review and add the static analysis checks

drupol:
Sounds good :)💪👍

src/Contract/Operation/Spanable.php
```diff
@@ -18,7 +19,9 @@
 interface Spanable
 {
     /**
-     * @return Collection<TKey, T>
+     * @param callable(T, TKey, Iterator<TKey, T>): bool $callback
```
AlexandruGG, line 22: shouldn't `Span` also accept variadic callable?

AlexandruGG:
@drupol this should be ready once the mutation tests finish (the Scrutinizer failure seems unrelated). I polished the documentation a bit, added a few more tests, and added the static analysis checks

src/Contract/Operation/Spanable.php
drupol: Nice catch, indeed it could be!

drupol: But it could be tackled in a new PR.

drupol:
All good, thanks !

AlexandruGG:
> All good, thanks !

Nice, we're gonna have another big release soon 😄

drupol:
Ho yes :)

PR #125, opened 2021-07-06
Add `Reject` operation.

drupol:
This PR

* [x] Add a new operation Reject.
* [x] Has tests
* [x] Has documentation
* [x] Has static analysis tests

Depends on #126 
Related to #123

### DIFF
--- a/docs/pages/code/operations/reject.php
+++ b/docs/pages/code/operations/reject.php
@@ -1,0 +1,24 @@
+<?php
+
+/**
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace App;
+
+use loophp\collection\Collection;
+
+include __DIR__ . '/../../../../vendor/autoload.php';
+
+$divisibleBy2 = static fn ($value): bool => 0 === $value % 2;
+$divisibleBy3 = static fn ($value): bool => 0 === $value % 3;
+$divisibleBy4 = static fn ($value): bool => 0 === $value % 4;
+
+$collection = Collection::fromIterable(range(1, 10))
+    ->reject($divisibleBy2); // [1, 3, 5, 7, 9]
+
+$collection = Collection::fromIterable(range(1, 10))
+    ->reject($divisibleBy2, $divisibleBy3, $divisibleBy4); // [1, 5, 7]
--- a/docs/pages/code/operations/reject.php
+++ b/docs/pages/code/operations/reject.php
@@ -15,10 +15,13 @@ include __DIR__ . '/../../../../vendor/autoload.php';
 
 $divisibleBy2 = static fn ($value): bool => 0 === $value % 2;
 $divisibleBy3 = static fn ($value): bool => 0 === $value % 3;
-$divisibleBy4 = static fn ($value): bool => 0 === $value % 4;
 
 $collection = Collection::fromIterable(range(1, 10))
     ->reject($divisibleBy2); // [1, 3, 5, 7, 9]
 
 $collection = Collection::fromIterable(range(1, 10))
-    ->reject($divisibleBy2, $divisibleBy3, $divisibleBy4); // [1, 5, 7]
+    ->reject($divisibleBy2, $divisibleBy3); // [1, 5, 7]
+
+$collection = Collection::fromIterable(range(1, 10))
+    ->reject($divisibleBy2)
+    ->reject($divisibleBy3); // [1, 5, 7]
--- a/src/Operation/Reject.php
+++ b/src/Operation/Reject.php
@@ -1,0 +1,95 @@
+<?php
+
+/**
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace loophp\collection\Operation;
+
+use Closure;
+use Generator;
+use Iterator;
+
+/**
+ * @immutable
+ *
+ * @template TKey
+ * @template T
+ *
+ * phpcs:disable Generic.Files.LineLength.TooLong
+ */
+final class Reject extends AbstractOperation
+{
+    /**
+     * @pure
+     *
+     * @return Closure(callable(T, TKey, Iterator<TKey, T>):bool...): Closure(Iterator<TKey, T>): Generator<int, list<array{0: TKey, 1: T}>>
+     */
+    public function __invoke(): Closure
+    {
+        return
+            /**
+             * @param callable(T, TKey, Iterator<TKey, T>): bool ...$callbacks
+             *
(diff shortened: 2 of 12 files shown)


AlexandruGG:
@drupol if you solve the conflicts here I can give this one a good look as well

drupol:
Rebased and ready to review!

tests/static-analysis/reject.php
```diff
@@ -35,14 +35,14 @@ function reject_checkMap(CollectionInterface $collection): void
...
-reject_checkList(Collection::fromIterable([1, 2, 3])->filter($intValueCallback, $intValueCallback2));
-reject_checkList(Collection::fromIterable([1, 2, 3])->filter($intKeyValueCallback1));
-reject_checkList(Collection::fromIterable([1, 2, 3])->filter($intKeyValueCallback1, $intKeyValueCallback2));
-
-reject_checkMap(Collection::fromIterable(['foo' => 'bar', 'baz' => ''])->filter());
-reject_checkMap(Collection::fromIterable(['foo' => 'bar', 'baz' => ''])->filter($stringValueCallback));
-reject_checkMap(Collection::fromIterable(['foo' => 'bar', 'baz' => ''])->filter($stringValueCallback, $stringValueCallback2));
-reject_checkMap(Collection::fromIterable(['foo' => 'bar', 'baz' => ''])->filter($stringKeyValueCallback1));
-reject_checkMap(Collection::fromIterable(['foo' => 'bar', 'baz' => ''])->filter($stringKeyValueCallback1, $stringKeyValueCallback2));
+reject_checkList(Collection::fromIterable([1, 2, 3])->reject());
```
drupol, line 38: #facepalm ... I should more careful sometimes !

src/Operation/Reject.php
```diff
@@ -26,7 +26,7 @@ final class Reject extends AbstractOperation
     /**
      * @pure
      *
-     * @return Closure(callable(T, TKey, Iterator<TKey, T>):bool...): Closure(Iterator<TKey, T>): Generator<int, list<array{0: TKey, 1: T}>>
+     * @return Closure(callable(T , TKey, Iterator<TKey, T>): bool ...): Closure (Iterator<TKey, T>): Generator<TKey, T>
```
drupol, line 29: Ok I won't complain about extra spaces, but we should fix them.

AlexandruGG: no you're right there were a couple that were wrong, fixed now

docs/pages/code/operations/reject.php
```diff
@@ -0,0 +1,27 @@
...
+
+$collection = Collection::fromIterable(range(1, 10))
+    ->reject($divisibleBy2); // [1, 3, 5, 7, 9]
+
+$collection = Collection::fromIterable(range(1, 10))
+    ->reject($divisibleBy2, $divisibleBy3); // [1, 5, 7]
+
+$collection = Collection::fromIterable(range(1, 10))
+    ->reject($divisibleBy2)
+    ->reject($divisibleBy3); // [1, 5, 7]
```
AlexandruGG, line 27: is it just me or because of the nature of `reject`, calling it with variadic callbacks is the same as calling it multiple times? Acts as logical `AND`

AlexandruGG (commented):
This is missing tests 😉

docs/pages/code/operations/reject.php
drupol: Yes exactly !

Not "A or B" is "Not A and Not B"

src/Operation/Reject.php
drupol: Thanks :)

drupol:
> This is missing tests wink

Going to do this tomorrow, I'm out for today.

AlexandruGG:
> This is missing tests 😉

On this, I was surprised nothing in the CI workflows alerted us about it. But now I see there's nothing in the workflows to cause it to fail if a certain code coverage % is not reached. Would you want to have something like that?

Another two things I noticed looking into the jobs 😄

1. this is broken

Screenshot 2021-07-06 at 22 53 34


2. we are running the jobs twice, it would be sufficient to either trigger them on `pull_request` OR on `push`
Screenshot 2021-07-06 at 22 54 30


I can create a separate PR to fix number 2, but not too sure about number 1

drupol:
> > This is missing tests wink
> 
> On this, I was surprised nothing in the CI workflows alerted us about it. But now I see there's nothing in the workflows to cause it to fail if a certain code coverage % is not reached. Would you want to have something like that?

Yes good idea and it's quite easy to do, it's already done here: https://github.com/ecphp/cas-bundle/blob/master/grumphp.yml.dist#L9
 
> Another two things I noticed looking into the jobs smile
> 
>     1. this is broken
> 
> 
> Screenshot 2021-07-06 at 22 53 34

Yes this is broken on Windows, and it's "normal". This step is only supposed to be triggered on Linux. This is also why `continue-on-error: true` is there.
 
>     1. we are running the jobs twice, it would be sufficient to either trigger them on `pull_request` OR on `push`
> 
> 
> Screenshot 2021-07-06 at 22 54 30

Yeah I think that could be fixed in a separate PR indeed.

AlexandruGG:
> > On this, I was surprised nothing in the CI workflows alerted us about it. But now I see there's nothing in the workflows to cause it to fail if a certain code coverage % is not reached. Would you want to have something like that?
> 
> Yes good idea and it's quite easy to do, it's already done here: https://github.com/ecphp/cas-bundle/blob/master/grumphp.yml.dist#L9

on this, does it mean that the coverage check is already enforced but the level (80%) is too low so that's why it didn't fail?

drupol:
> > > On this, I was surprised nothing in the CI workflows alerted us about it. But now I see there's nothing in the workflows to cause it to fail if a certain code coverage % is not reached. Would you want to have something like that?
> > 
> > 
> > Yes good idea and it's quite easy to do, it's already done here: https://github.com/ecphp/cas-bundle/blob/master/grumphp.yml.dist#L9
> 
> on this, does it mean that the coverage check is already enforced but the level (80%) is too low so that's why it didn't fail?

That's for ecphp/cas-bundle project :smile:  !

AlexandruGG (approved):

AlexandruGG:
> > > > On this, I was surprised nothing in the CI workflows alerted us about it. But now I see there's nothing in the workflows to cause it to fail if a certain code coverage % is not reached. Would you want to have something like that?
> > > 
> > > 
> > > Yes good idea and it's quite easy to do, it's already done here: https://github.com/ecphp/cas-bundle/blob/master/grumphp.yml.dist#L9
> > 
> > 
> > on this, does it mean that the coverage check is already enforced but the level (80%) is too low so that's why it didn't fail?
> 
> That's for ecphp/cas-bundle project 😄 !

Oops misread it, I thought it was on the php-conventions project 😄

PR #126, opened 2021-07-06
Update `Filter` operation.

drupol:
All the operations having variadic arguments evaluates as logical `OR`, the filter operation should also follow that rule.

**BREAKING CHANGE: yes**

This PR

* [x] Update the Filter operation
* [x] Has updated tests

Related to #123 and #125

### DIFF
--- a/src/Operation/Filter.php
+++ b/src/Operation/Filter.php
@@ -9,8 +9,8 @@ declare(strict_types=1);
 
 namespace loophp\collection\Operation;
 
-use CallbackFilterIterator;
 use Closure;
+use Generator;
 use Iterator;
 
 /**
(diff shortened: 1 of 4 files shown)


src/Operation/Filter.php
```diff
@@ -26,23 +26,49 @@ final class Filter extends AbstractOperation
...
             static fn (callable ...$callbacks): Closure =>
                 /**
                  * @param Iterator<TKey, T> $iterator
                  *
-                 * @return Iterator<TKey, T>
+                 * @return Generator<TKey, T>
                  */
-                static function (Iterator $iterator) use ($callbacks): Iterator {
+                static function (Iterator $iterator) use ($callbacks): Generator {
+                    // TODO: Find a way to avoid repeating this everywhere.
```
AlexandruGG, line 46: I just noticed this pattern in the other operations now that you commented this 😄.

This can't be an operation by itself so where could we put it 🤔. Maybe in `AbstractOperation` or as a trait?

drupol: Dunno yet, but it's going to be a follow up for sure.

AlexandruGG (approved):
Given the BC break is this going out in 4.2?

src/Operation/Filter.php
AlexandruGG: I was going to bring up the fact that we should aim to look at some of the TODOs in the codebase 😁

PR #127, opened 2021-07-07
Change GitHub Actions Workflow Triggers

AlexandruGG:
This PR looks to solve workflows triggering multiple times on PRs.

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -1,61 +1,64 @@
 # https://help.github.com/en/categories/automating-your-workflow-with-github-actions
 
 on:
-    - pull_request
+  push:
+    branches:
+      - master
+  pull_request:
 
 name: "Static analysis"
 
 jobs:
-    run:
-        name: "Static Analysis"
-        runs-on: ${{ matrix.operating-system }}
-        strategy:
-            fail-fast: false
-            matrix:
-                operating-system: [ubuntu-latest]
-                php-versions: ['7.4', '8.0']
+  run:
+    name: "Static Analysis"
+    runs-on: ${{ matrix.operating-system }}
+    strategy:
+      fail-fast: false
+      matrix:
+        operating-system: [ubuntu-latest]
+        php-versions: ["7.4", "8.0"]
 
-        steps:
-            - name: Set git to use LF
-              run: |
-                  git config --global core.autocrlf false
-                  git config --global core.eol lf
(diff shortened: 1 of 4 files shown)


.github/workflows/static-analysis.yml
```diff
@@ -2,7 +2,6 @@
 
 on:
     - pull_request
-    - push
```
AlexandruGG, line 5: @drupol do you want these to run in `master` as well, after a merge?

Currently we have them set up like that but I don't know if it was intentional or not: https://github.com/loophp/collection/actions?query=branch%3Amaster

drupol: I don't think it's necessary. All the changes happening in this project should be committed through PRs and not directly in the master branch, so... it will be tested in PRs. WDYT ?

AlexandruGG: Screenshot 2021-07-07 at 09 22 07

I think some of the badges on the readme depend on that 🤔

.github/workflows/tests.yml
```diff
@@ -1,57 +1,59 @@
 # https://help.github.com/en/categories/automating-your-workflow-with-github-actions
 
 on:
-    - pull_request
-    - push
+  push:
+    branches:
+      - master
+  pull_request:
```
AlexandruGG, line 7: I decided to keep it on the safe side so this means it runs:
- on pull requests
- on pushes but only to the `master` branch

drupol (changes requested):
The indentation has been changed... and it was not supposed to.

Based on the `.editorconfig`, the indentation is 4 spaces.

If you're using VSCode or VSCodium with the `.editorconfig` plugin, do: `SHIFT+CTRL+i` to reformat the file correctly.

.github/workflows/static-analysis.yml
drupol: Fair enough, let's not modify these.

AlexandruGG:
> The indentation has been changed... and it was not supposed to.
> 
> Based on the `.editorconfig`, the indentation is 4 spaces.
> 
> If you're using VSCode or VSCodium with the `.editorconfig` plugin, do: `SHIFT+CTRL+i` to reformat the file correctly.

Nice, thanks for that I wasn't using the `.editorconfig` plugin but I got it now 😁 . Formatted with it now.

drupol:
Merci :)

PR #130, opened 2021-07-07
Add Collection::isEmpty Method

AlexandruGG:
### Overview
This PR adds the `isEmpty` method and `IsEmptyable` interface.

**Why is this needed when we have `Collection::count`?**

### 1. Convenience
Typical use case:
```PHP
$collection = Collection::fromIterable([1, 2, 3]);
$filteredCollection = $collection->filter(...);

if (!$filteredCollection->isEmpty()) {
    // do something
}

// compared to

if (0 !== $filteredCollection->count()) {
    // do something
}
```

Other popular PHP collection libraries also have this, so people are used to it:
**Doctrine**: https://www.doctrine-project.org/projects/doctrine-collections/en/1.6/index.html#isempty
**Laravel**: https://laravel.com/docs/8.x/collections#method-isempty
**Knapsack**: https://dusankasan.github.io/Knapsack/#Operations-isEmpty

Other languages have this:
**Java**: https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#isEmpty--
**C++**: https://www.cplusplus.com/reference/vector/vector/empty/
**Haskell**: http://zvon.org/other/haskell/Outputprelude/null_f.html

### 2. Performance
This is a big one. Simple benchmark using `microtime` and `Collection::range(0, 100_000_000)`:

```PHP
$isEmpty = 0 === $collection->count();
// vs
$isEmpty = $collection->isEmpty();
```

Ran it 100,000 times for `isEmpty()`, only 10 times for `count`, and here's why (average time for _a single run_):

 `isEmpty` -> **5.575289726257324E-6 seconds** (basically instant because it's in constant time)
 `count` -> **13.571631002426148 seconds** (this runs in O(N))

### Checklist

- [x] Tests
- [x] Static analysis check
- [x] Documentation

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -476,9 +476,12 @@ Signature: ``Collection::contains(...$values);``
     }
 
 count
-~~~~~~~~
+~~~~~
 
-Returns the number of elements in a collection
+Returns the number of elements in a collection.
+
+.. tip:: If you only want to check whether the collection is empty or not, use the ``isEmpty`` 
+         operation as it will be significant more performant in a large collection.
 
 Interface: `Countable`_
 
@@ -1170,6 +1173,26 @@ Signature: ``Collection::intersperse($element, int $every = 1, int $startAt = 0)
     foreach($collection as $item) {
         var_dump($item); // 'x', 'a', 'x', 'b', 'x', 'c'
     }
+
+isEmpty
+~~~~~~~
+
+Check if a collection has any elements inside.
+
+Interface: `IsEmptyable`_
+
+Signature: ``Collection::isEmpty();``
+
+.. code-block:: php
+
+    Collection::fromIterable(range('a', 'c'))
+        ->isEmpty(); // false
+
+    Collection::fromIterable([])
+        ->isEmpty(); // true
+
+    Collection::fromIterable([null])
+        ->isEmpty(); // false
 
 key
 ~~~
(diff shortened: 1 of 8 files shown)


src/Collection.php
```diff
@@ -562,6 +562,11 @@ public function intersperse($element, int $every = 1, int $startAt = 0): Collect
         return new self(Intersperse::of()($element)($every)($startAt), $this->getIterator());
     }
 
+    public function isEmpty(): bool
+    {
+        return !$this->getIterator()->valid();
+    }
```
AlexandruGG, line 568: I got the idea for it from Laravel's LazyCollection: https://github.com/illuminate/collections/blob/master/LazyCollection.php#L556

AlexandruGG:
@drupol I hope you'll like this one 😁

docs/pages/api.rst
```diff
@@ -1171,6 +1174,26 @@ Signature: ``Collection::intersperse($element, int $every = 1, int $startAt = 0)
...
+
+Check if a collection has any elements inside.
+
+Interface: `IsEmptyable`_
+
+Signature: ``Collection::isEmpty();``
+
+.. code-block:: php
+
+    Collection::fromIterable(range('a', 'c'))
```
drupol, line 1188: One minor remark. How about moving the example in its own file like we did for others ?

src/Collection.php
drupol: Yep, this is how it is supposed to be done.

drupol (changes requested):
I'm ok for it just a minor thing to do and we are good.

docs/pages/api.rst
drupol: I just did it so it's ok.

drupol (approved):

drupol:
Thanks, good addition !

PR #131, opened 2021-07-08
Provide util class for reducer callback

drupol:
This PR

* [x] Provide a `ReducerBuilder` functional and stateless class
* [x] Replace the use of custom reducer builder by using the new `ReducerBuilder` class. 
* [x] Has tests

### DIFF
--- a/spec/loophp/collection/Utils/CallbacksArrayReducerSpec.php
+++ b/spec/loophp/collection/Utils/CallbacksArrayReducerSpec.php
@@ -10,26 +10,45 @@ declare(strict_types=1);
 namespace spec\loophp\collection\Utils;
 
 use ArrayIterator;
+use Iterator;
 use loophp\collection\Utils\CallbacksArrayReducer;
 use PhpSpec\ObjectBehavior;
 
 class CallbacksArrayReducerSpec extends ObjectBehavior
 {
+    public function it_ensure_callbacks_receive_the_needed_arguments(): void
+    {
+        $callbacks = [
+            static fn (string $value, string $key, Iterator $iterator): bool => 'value_key_a_b_c' === sprintf('%s_%s_%s', $value, $key, implode('_', iterator_to_array($iterator))),
+        ];
+
+        $iterator = new ArrayIterator(range('a', 'c'));
+
+        $this::or()($callbacks, 'value', 'key', $iterator)
+            ->shouldReturn(true);
+    }
+
     public function it_is_initializable()
     {
         $this->shouldHaveType(CallbacksArrayReducer::class);
     }
 
-    public function it_reduce_callbacks(): void
+    public function it_reduces_empty_callbacks_array(): void
+    {
+        $iterator = new ArrayIterator(range(0, 10));
+
+        $this::or()([], 0, 0, $iterator)
+            ->shouldReturn(false);
+    }
+
+    public function it_reduces_multiple_callbacks(): void
     {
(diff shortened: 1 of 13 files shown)


src/Operation/Reject.php
```diff
@@ -80,11 +55,7 @@ static function (Iterator $iterator) use ($callbacks): Generator {
                         $callbacks;
 
                     foreach ($iterator as $key => $current) {
-                        $result = array_reduce(
-                            $callbacks,
-                            $reducerCallback($key)($current)($iterator),
-                            false
-                        );
+                        $result = OrCallbackReducer::or()($callbacks, $current, $key, $iterator);
```
AlexandruGG, line 58: I think having `Or` twice in this is a bit redundant.

I propose either:

```suggestion
                        $result = CallbackReducer::or()($callbacks, $current, $key, $iterator);
```

this leaves room to have `CallbackReducer::and()`

OR 

```suggestion
                        $result = OrCallbackReducer::of()($callbacks, $current, $key, $iterator);
```

and we make it invokable

drupol: Fair enough, done !

spec/loophp/collection/Utils/CallbacksArrayReducerSpec.php
```diff
@@ -0,0 +1,43 @@
...
+
+        $this::or()($callbacks, 0, 0, $iterator)
+            ->shouldReturn(true);
+
+        $this::or()($callbacks, 13, 0, $iterator)
+            ->shouldReturn(true);
+
+        $this::or()($callbacks, 3, 0, $iterator)
+            ->shouldReturn(false);
+    }
```
AlexandruGG, line 42: A few comments here:
- the `$iterator` is not really used in this test because none of the callbacks use it. Would be good to have one test which uses it
- similarly, would be good to have a test with callbacks that use the second argument (the key)
- we should also test some edge cases, e.g. empty callbacks array or a single callback
- it's good practice to have one assertion per test. I know with `CollectionSpec` maybe that's not too convenient because it's a big class but with something like this I think it makes it easier to follow, e.g.
`it_reduces_empty_callbacks_array`
`it_reduces_single_callback`
`it_reduces_callbacks_with_key_and_value`
`it_reduces_callbacks_with_all_parameters`

drupol: Ok, will add this tonight !

drupol: Job done!

AlexandruGG (approved):
Nice one 🎉

src/Utils/CallbacksArrayReducer.php
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * @immutable
+ *
+ * @template TKey
+ * @template T
+ *
+ * phpcs:disable Generic.Files.LineLength.TooLong
+ */
+final class CallbacksArrayReducer
```
AlexandruGG, line 23: We could shorten the name to `CallbacksReducer`. Due to the plural it's quite clear we are dealing with a list of multiple callbacks

drupol:
Thanks :+1:

src/Utils/CallbacksArrayReducer.php
drupol: Oops, too late, we can tackle that in a follow-up PR...

PR #132, opened 2021-07-09
Refactor Scalar Operations

AlexandruGG:
This PR refactors Collection Operations which currently return a single scalar value inside a collection, returning the scalar directly.

* [x] Contains 
* [x] Every
* [x] Falsy
* [x] Has 
* [x] Match + static analysis check & type fixes
* [x] Nullsy
* [x] Truthy

Related to https://github.com/loophp/collection/discussions/107.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ This library has been inspired by:
     needed are standard PHP variables like `int`, `string`, `callable` or
     `iterator`.
     It allows users to use those operations individually, at their own will, to
-    build up something custom. Currently, more than [**80 operations**][28] are
+    build up something custom. Currently, more than [**100 operations**][28] are
     available in this library. This library is an example of what you can do
     with all those small bricks, but nothing prevents users from using an operation
     on its own as well.
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -2072,14 +2072,10 @@ Signature: ``Collection::truthy();``
 .. code-block:: php
 
     $truthyCollection = Collection::fromIterable([2, 3, 4])
-        ->truthy(); // [true]
+        ->truthy(); // true
 
     $falsyCollection = Collection::fromIterable(['a', '', 'c', 'd'])
-        ->truthy(); // [false]
-
-    if ($falsyCollection->truthy()->current()) {
-        // do something
-    }
+        ->truthy(); // false
 
 unlines
 ~~~~~~~
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -452,8 +452,7 @@ contains
 
 Check if the collection contains one or more values.
 
-.. warning:: The `values` parameter is variadic and will be evaluated as a logical ``OR``.
-             If you're looking for a logical ``AND``, you have to make separate calls to this method.
+.. warning:: The ``values`` parameter is variadic and will be evaluated as a logical ``OR``.
 
 Interface: `Containsable`_
 
@@ -647,9 +646,7 @@ every
 
 This operation tests whether all elements in the collection pass the test implemented by the provided callback(s).
 
-.. warning:: The `callbacks` parameter is variadic and will be evaluated as a logical ``OR``.
-             If you're looking for a logical ``AND``, you have to make multiple calls to the
-             same operation.
+.. warning:: The ``callbacks`` parameter is variadic and will be evaluated as a logical ``OR``.
 
 Interface: `Everyable`_
 
@@ -1000,8 +997,6 @@ has
 Check if the collection has values with the help of one or more callables.
 
 .. warning:: The ``callbacks`` parameter is variadic and will be evaluated as a logical ``OR``.
-             If you're looking for a logical ``AND``, you have to make multiple calls to the
-             same operation.
 
 Interface: `Hasable`_
 
@@ -1017,7 +1012,7 @@ Signature: ``Collection::has(callable ...$callbacks);``
 
     Collection::fromIterable(range('A', 'C'))
         ->has(
-            static fn ($value, $key): string => $key > 4 ? 'D' : A',
+            static fn ($value, $key): string => $key > 4 ? 'D' : 'A',
             static fn ($value, $key): string => 'Z'
         ); // true
 
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -160,7 +160,7 @@ When used as a ``Collection`` method, operations fall into three main categories
 1. Operations that return a ``scalar`` value. Currently, this includes:
    ``Contains``, ``Every``, ``Falsy``, ``Has``, ``Key``, ``Match`` (or ``MatchOne``), ``Nullsy``, ``Truthy``.
 
-2. Operations that return a new ``Collection`` object - this includes the majority of operations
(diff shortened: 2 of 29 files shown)


docs/pages/api.rst
```diff
@@ -453,8 +453,7 @@ contains
 Check if the collection contains one or more values.
 
 .. warning:: The `values` parameter is variadic and will be evaluated as a logical ``OR``.
-             If you're looking for a logical ``AND``, you have to make separate calls to this method;
-             note the calls cannot be in succession because the collection will contain a boolean after the first call.
+             If you're looking for a logical ``AND``, you have to make separate calls to this method.
```
drupol, line 456: I think we can get rid of this line, it doesn't make sense actually.

AlexandruGG: you're right, I simplified this for the other operations that return booleans. It's quite obvious that separate calls should be made if you're trying to test multiple things as an `AND` if the operation returns a boolean 😄

src/Contract/Operation/Matchable.php
```diff
@@ -20,9 +19,7 @@ interface Matchable
 {
     /**
      * @param callable(T, TKey, Iterator<TKey, T>): bool $callback
-     * @param null|callable(T, TKey, Iterator<TKey, T>): T $matcher
-     *
-     * @return Collection<int, bool>
+     * @param null|callable(T, TKey, Iterator<TKey, T>): bool $matcher
```
AlexandruGG, line 22: this is now correct I believe? both the callback and the matcher should return booleans

AlexandruGG:
I'm gonna give this one a final look a bit later before marking it ready for review

src/Contract/Operation/Matchable.php
drupol: Yes that's how it supposed to work.

drupol:
> I'm gonna give this one a final look a bit later before marking it ready for review

Take your time mate, there is no hurry.

Do you think it would be possible to add some documentation that explains

* All the methods of the Collection returns a new collection
* Only the methods X,Y,Z returns something different
* Explaining the reasons of this breaking change.

I will also use that part in the changelog when doing the release later on.

AlexandruGG:
> > I'm gonna give this one a final look a bit later before marking it ready for review
> 
> Take your time mate, there is no hurry.
> 
> Do you think it would be possible to add some documentation that explains
> 
> * All the methods of the Collection returns a new collection
> * Only the methods X,Y,Z returns something different
> * Explaining the reasons of this breaking change.
> 
> I will also use that part in the changelog when doing the release later on.

Yeah for sure! I was thinking to add a part in the documentation to explain about the operations but I forgot about it there for a moment 😄.

Regarding the last bullet point do you want that to be in the documentation or do you mean in the PR description?

drupol:
Ok good 👍

I think we can do it in the documentation and I'll copy paste it for the changelog later.

AlexandruGG:
> Ok good 👍
> 
> I think we can do it in the documentation and I'll copy paste it for the changelog later.

Sounds good. Btw for BC breaking changes like this should we have a new major version? I.e 5.0?

drupol:
I don't know yet. We will see when it's done.

README.md
```diff
@@ -76,7 +76,7 @@ This library has been inspired by:
     needed are standard PHP variables like `int`, `string`, `callable` or
     `iterator`.
     It allows users to use those operations individually, at their own will, to
-    build up something custom. Currently, more than [**80 operations**][28] are
+    build up something custom. Currently, more than [**100 operations**][28] are
```
AlexandruGG, line 79: did a new count and this is up over 100 now 🎉

AlexandruGG:
> I don't know yet. We will see when it's done.

I think it depends if we want this project to follow [semver](https://semver.org/) or not. Because based on that any backwards-incompatible API changes should be done as part of a new **major** version.

Also if we do up the version when releasing this we should up the [docs](https://github.com/loophp/collection/blob/master/docs/conf.py#L60) version.

AlexandruGG:
@drupol let me know what you think about the documentation update

README.md
drupol: Excellent :-)

docs/pages/api.rst
```diff
@@ -145,7 +145,32 @@ Another example
...
+or through the ``Collection`` object.
+
+.. literalinclude:: code/operations/background.php
+  :language: php
+
+When used separately, operations typically return a PHP `Generator`_ or an `Iterator`_.
+When used as a ``Collection`` method, operations fall into three main categories based on the return type:
+
+1. Operations that return a ``scalar`` value. Currently, this includes:
+   ``Contains``, ``Every``, ``Falsy``, ``Has``, ``Key``, ``Match`` (or ``MatchOne``), ``Nullsy``, ``Truthy``.
```
drupol, line 161: I think the explanation is overall really great and I wouldn't change a thing, except for the classification points here.

What I propose is to list the operation types as such:

1. Operations which returns a boolean: countains, every, falsy, has... etc etc
2. Operations which returns a Collection of Collection: span, partition
3. Operation which returns an array: all
4. Operation which returns a Collection: all the rest.

(_be carefull _Collection::key()_ might return anything, not only a scalar, but anything_)

Also, what do you think about:

* Updating the signatures of all the operations in the API page and add their return type ?
or
* Updating all operations and add a note about the return type ?

(_about that page, it's very sad we cannot generate the documentation automatically from the classes themselves!_)

Let me know what you think.

drupol (changes requested):

drupol:
> > I don't know yet. We will see when it's done.
> 
> I think it depends if we want this project to follow [semver](https://semver.org/) or not. Because based on that any backwards-incompatible API changes should be done as part of a new **major** version.
> 
> Also if we do up the version when releasing this we should up the [docs](https://github.com/loophp/collection/blob/master/docs/conf.py#L60) version as well.

We should definitely follow semver... and therefore it's going to be version 5 indeed.

AlexandruGG:
> > > I don't know yet. We will see when it's done.
> > 
> > 
> > I think it depends if we want this project to follow [semver](https://semver.org/) or not. Because based on that any backwards-incompatible API changes should be done as part of a new **major** version.
> > Also if we do up the version when releasing this we should up the [docs](https://github.com/loophp/collection/blob/master/docs/conf.py#L60) version as well.
> 
> We should definitely follow semver... and therefore it's going to be version 5 indeed.

Cool in that case I think we can do the `map` change as well to remove the [deprecation](https://github.com/loophp/collection/blob/master/src/Operation/Map.php#L48-L53) and only allow one callback, I'll do it in a future PR.

docs/pages/api.rst
AlexandruGG: @drupol thank you for the feedback, I've done some changes to the numbered list to incorporate your suggestions, with a couple tweaks. Let me know what you think of this new classification :).

Regarding updating the signatures of the operations in the docs -> it's something that I thought about today as well 😄 . I'll do it in a future PR for all of them

drupol: Thanks all good now :)

AlexandruGG: @drupol great, this is all ready on my side 😁

drupol:
Thanks, a big one done :1st_place_medal:

PR #133, opened 2021-07-11
Add All Operation

AlexandruGG:
This PR:

* [x] Adds an `All` operation
* [x] Has documentation

Even though this is a simple one, it's good to have it in an operation so that users that want to use the operations separately can benefit from it and don't have to use `iterator_to_array` all the time.

### DIFF
--- a/spec/loophp/collection/CollectionSpec.php
+++ b/spec/loophp/collection/CollectionSpec.php
@@ -27,6 +27,7 @@ use PhpSpec\Exception\Example\FailureException;
 use PhpSpec\Exception\Example\MatcherException;
 use PhpSpec\ObjectBehavior;
 use stdClass;
+use TypeError;
 use function gettype;
 use const E_USER_DEPRECATED;
 use const INF;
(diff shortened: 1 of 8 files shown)


spec/loophp/collection/CollectionSpec.php
```diff
@@ -27,13 +27,53 @@
...
+
+            yield ['b'] => 2;
+        };
+
+        $this::fromIterable($nonArrayKeyGen())
+            ->shouldIterateAs($nonArrayKeyGen());
+
+        $this::fromIterable($nonArrayKeyGen())
+            ->shouldThrow(TypeError::class)
+            ->during('all');
```
AlexandruGG, line 74: I actually didn't realise this before but `all` cannot be used at all if you have keys that are not `int|string`, a type error will be thrown

src/Operation/Pipe.php
```diff
@@ -26,31 +24,31 @@ final class Pipe extends AbstractOperation
     /**
      * @pure
      *
-     * @return Closure(...callable(Iterator<TKey, T>):Generator<TKey, T, mixed, mixed>):Closure(Iterator<TKey, T>):Iterator<TKey, T>
+     * @return Closure(callable(iterable<TKey, T>): iterable<TKey, T> ...): Closure(iterable<TKey, T>): iterable<TKey, T>
```
AlexandruGG, line 27: I expanded the type here from `Iterator` to `iterable` to allow for the `All` usage, which returns an `array`

drupol: Nice one!

drupol:
Thanks :dancing_men:

PR #136, opened 2021-07-11
Update Operations Signatures in API Docs

AlexandruGG:
This PR updates Collection Operations signatures in the `api` docs, adding types and correcting a few outdated ones.


AlexandruGG:
@drupol 

> Updating the signatures of all the operations in the API page and add their return type ?

sorted 😁

drupol:
Amazing work!

Thanks for also fixing the other things as well along the way!

PR #139, opened 2021-07-11
Add `Reduce` operation

drupol:
This PR:

* [x] Provide a new operation: `Reduce`
* [x] Has documentation
* [x] Have proper static typing

### DIFF
--- a/docs/pages/code/operations/reduce.php
+++ b/docs/pages/code/operations/reduce.php
@@ -19,4 +19,4 @@ $collection = Collection::empty()
     ->reduce($callback); // []
 
 $collection = Collection::fromIterable(range(1, 5))
-    ->reduce($callback); // [15]
+    ->reduce($callback, 0); // [4 => 15]
--- a/docs/pages/code/operations/reduce.php
+++ b/docs/pages/code/operations/reduce.php
@@ -1,0 +1,22 @@
+<?php
+
+/**
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace App;
+
+use loophp\collection\Collection;
+
+include __DIR__ . '/../../../../vendor/autoload.php';
+
+$callback = static fn (int $carry, int $item): int => $carry + $item;
+
+$collection = Collection::empty()
+    ->reduce($callback); // []
+
+$collection = Collection::fromIterable(range(1, 5))
+    ->reduce($callback); // [15]
--- a/spec/loophp/collection/CollectionSpec.php
+++ b/spec/loophp/collection/CollectionSpec.php
@@ -1425,6 +1425,10 @@ class CollectionSpec extends ObjectBehavior
 
     public function it_can_fold_from_the_left(): void
     {
+        $this::empty()
+            ->foldLeft(static fn (string $carry, string $string): string => sprintf('%s%s', $carry, $string), 'foo')
+            ->shouldIterateAs(['foo']);
+
         $this::fromIterable(range('A', 'C'))
             ->foldLeft(
                 static function (string $carry, string $item): string {
@@ -2648,6 +2652,29 @@ class CollectionSpec extends ObjectBehavior
             ->random(0)
             ->shouldThrow(OutOfBoundsException::class)
             ->during('all');
+    }
+
+    public function it_can_reduce(): void
+    {
+        $this::empty()
+            ->reduce(static fn (string $carry, string $string): string => sprintf('%s%s', $carry, $string), 'foo')
+            ->shouldIterateAs([]);
+
+        $this::fromIterable(range(1, 5))
+            ->reduce(
+                static fn (int $carry, int $item): int => $carry + $item,
+                0
+            )
+            ->shouldIterateAs([4 => 15]);
+
+        $this::fromIterable(array_combine(range('x', 'z'), range('a', 'c')))
+            ->reduce(
+                static fn (string $carry, string $letter, string $index): string => sprintf('%s[%s:%s]', $carry, $index, $letter),
+                '=> '
+            )
+            ->shouldIterateAs([
+                'z' => '=> [x:a][y:b][z:c]',
+            ]);
     }
 
     public function it_can_reduction(): void
--- a/src/Contract/Operation/Reduceable.php
+++ b/src/Contract/Operation/Reduceable.php
@@ -1,0 +1,32 @@
+<?php
+
+/**
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace loophp\collection\Contract\Operation;
+
+use Iterator;
+use loophp\collection\Contract\Collection;
+
+/**
+ * @template TKey
+ * @template T
+ */
+interface Reduceable
+{
+    /**
+     * Reduce a collection of items through a given callback.
+     *
+     * @template V
+     *
+     * @param callable(V, T, TKey, Iterator<TKey, T>): V $callback
+     * @param V $initial
+     *
+     * @return Collection<TKey, V>
+     */
+    public function reduce(callable $callback, $initial): Collection;
+}
--- a/src/Contract/Operation/Reduceable.php
+++ b/src/Contract/Operation/Reduceable.php
@@ -28,5 +28,5 @@ interface Reduceable
      *
      * @return Collection<TKey, V>
      */
-    public function reduce(callable $callback, $initial): Collection;
+    public function reduce(callable $callback, $initial = null): Collection;
 }
--- a/src/Operation/GroupBy.php
+++ b/src/Operation/GroupBy.php
@@ -73,7 +73,7 @@ final class GroupBy extends AbstractOperation
 
                 /** @var Closure(Iterator<TKey, T>): Generator<int, list<T>> $pipe */
                 $pipe = Pipe::of()(
-                    FoldLeft::of()($reducerFactory($callable))([]),
+                    Reduce::of()($reducerFactory($callable))([]),
                     Flatten::of()(1)
                 );
 
--- a/tests/static-analysis/reduce.php
+++ b/tests/static-analysis/reduce.php
@@ -12,23 +12,24 @@ include __DIR__ . '/../../vendor/autoload.php';
 use loophp\collection\Collection;
 use loophp\collection\Contract\Collection as CollectionInterface;
 
-$sum = static fn (?int $carry, int $value): int => (int) $carry + $value;
-$concat = static fn (?string $carry, string $string): string => sprintf('%s%s', (string) $carry, $string);
-
 /**
  * @param CollectionInterface<int, int|null> $collection
+ */
+function reduce_checkListNullableInt(CollectionInterface $collection): void
+{
+}
+/**
+ * @param CollectionInterface<int, int> $collection
  */
 function reduce_checkListInt(CollectionInterface $collection): void
 {
 }
-
 /**
- * @param CollectionInterface<int, string|null> $collection
+ * @param CollectionInterface<string, string|null> $collection
  */
-function reduce_checkListString(CollectionInterface $collection): void
+function reduce_checkMapNullableString(CollectionInterface $collection): void
 {
 }
-
 /**
  * @param CollectionInterface<string, string> $collection
  */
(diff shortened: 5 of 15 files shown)


src/Contract/Operation/Reduceable.php
```diff
@@ -0,0 +1,28 @@
...
+interface Reduceable
+{
+    /**
+     * Reduce a collection of items through a given callback.
+     *
+     * @param mixed $initial
+     *
+     * @return Collection<TKey, T>
+     */
+    public function reduce(callable $callback, $initial = null): Collection;
```
AlexandruGG, line 27: This is not in sync with what's in the [operation](https://github.com/loophp/collection/pull/139/files?diff=split&w=1#diff-999d8f667f54a4a17936f748d9c02e4c5c9c727ef0181b9b714879554dd1ce16R31):
- missing type hint for the `callable`
- `$initial` should be the same type as the carry from the callback, right?

tests/static-analysis/reduce.php
```diff
@@ -0,0 +1,33 @@
...
+function reduce_checkListInt(CollectionInterface $collection): void
+{
+}
+
+/**
+ * @param CollectionInterface<int, string> $collection
+ */
+function reduce_checkListString(CollectionInterface $collection): void
+{
+}
```
AlexandruGG, line 30: rather than two list checks we should have a list and a "map" check, e.g. this should accept `CollectionInterface<string, string>`

spec/loophp/collection/CollectionSpec.php
```diff
@@ -2499,6 +2503,20 @@ public function it_can_random(): void
             ->during('all');
     }
 
+    public function it_can_reduce(): void
```
AlexandruGG, line 2506: would be good to have another test case where the collection is a map, e.g. `Collection::fromIterable(['foo' => 'f', 'bar' => 'b'])`

src/Contract/Operation/Reduceable.php
drupol: It's done, but I'm very unsure about this. I have the feeling that we cannot really properly type this.

tests/static-analysis/reduce.php
drupol: Done.

spec/loophp/collection/CollectionSpec.php
drupol: Done.

src/Operation/GroupBy.php
```diff
@@ -73,7 +73,7 @@ static function (array $collect, $value, $key) use ($callback): array {
 
                 /** @var Closure(Iterator<TKey, T>): Generator<int, list<T>> $pipe */
                 $pipe = Pipe::of()(
-                    FoldLeft::of()($reducerFactory($callable))([]),
+                    Reduce::of()($reducerFactory($callable))([]),
```
AlexandruGG, line 76: you mention in the commit message that this is more performant. Can you clarify why that is?

Also, maybe we could have this refactor in a dedicated PR?

drupol: If we could draw a tree of dependency for each Operation (_actually it could be a very nice idea_), we can see that

`FoldLeft`: [ [Reduction, Prepend], Last ]
`Reduce`: [Reduction, Last]

The `Reduce` operation does not need the `Prepend` operation and then, it's faster.

AlexandruGG: nice explanation, thanks. Indeed would be cool to have some sort of dependency tree.

I've used this before to enforce dependencies, but we can also use it to visualise them: https://github.com/qossmic/deptrac

drupol: I created the discussion to not forget this in the void! #149

drupol:
@AlexandruGG Do you think the static analysis stuff is good enough?

AlexandruGG:
> @AlexandruGG Do you think the static analysis stuff is good enough?

I'll give this a review tomorrow :)

tests/static-analysis/pair.php
```diff
@@ -0,0 +1,30 @@
...
+}
+
+/**
+ * @param CollectionInterface<string, string|null> $collection
+ */
+function pair_checkListString(CollectionInterface $collection): void
+{
+}
+
+pair_checkListInt(Collection::fromIterable([1, 2, 3])->pair());
```
AlexandruGG, line 29: this example specifically actually gives a PHP warning. I'm not sure if we can do anything to prevent this usage with the type hints, but probably out of scope for this PR

drupol: I don't have it locally, what's the warning ?

src/Contract/Operation/ScanLeftable.php
```diff
@@ -18,9 +19,12 @@
...
 {
     /**
-     * @param T|null $initial
+     * @template V
      *
-     * @return Collection<TKey, T>
+     * @param callable(V, T, TKey, Iterator<TKey, T>): V $callback
+     * @param V $initial
+     *
+     * @return Collection<int|TKey, T|V>
```
AlexandruGG, line 27:
```suggestion
     * @return Collection<TKey, V>
```
shouldn't it be like this? I'm not sure why we need to mention `int` separately, also it seems like the elements in the final collection will be of the type returned by the callback (`V`), just like in `Reduction`

src/Operation/ScanLeft.php
```diff
@@ -26,24 +26,26 @@ final class ScanLeft extends AbstractOperation
     /**
      * @pure
      *
-     * @return Closure(callable((T|null), T, TKey, Iterator<TKey, T>): (T|null)):Closure (T|null): Closure(Iterator<TKey, T>): Generator<int|TKey, T|null>
+     * @template V
+     *
+     * @return Closure(callable(V, T, TKey, Iterator<TKey, T>): V|T): Closure (V): Closure(Iterator<TKey, T>): Generator<TKey, V|T>
```
AlexandruGG, line 31: this is out of sync with the interface. Here the callable returns `V|T`, but in the interface it's `V`

src/Contract/Operation/ScanLeftable.php
drupol: Let me double-double-double-check :)

src/Contract/Operation/ScanRightable.php
```diff
@@ -18,9 +19,12 @@
...
 {
     /**
-     * @param T|null $initial
+     * @template V
      *
-     * @return Collection<TKey, T>
+     * @param callable(V, T, TKey, Iterator<TKey, T>): V $callback
+     * @param V $initial
+     *
+     * @return Collection<int|TKey, T|V>
```
AlexandruGG, line 27: same comment here

src/Contract/Operation/ScanLeftable.php
drupol: Nope, this is correct. The `ScanLeft` will return the initial value (V) if the `iterable` is empty, so, this is correct.

src/Operation/Reduction.php
```diff
@@ -26,27 +26,29 @@ final class Reduction extends AbstractOperation
     /**
      * @pure
      *
-     * @return Closure(callable((T|null), T, TKey, Iterator<TKey, T>): (T|null)):Closure (T|null): Closure(Iterator<TKey, T>): Generator<TKey, T>
+     * @template V
+     *
+     * @return Closure(callable(V, T, TKey, Iterator<TKey, T>): V): Closure (V): Closure(Iterator<TKey, T>): Generator<TKey, V>
```
AlexandruGG, line 31: can you add a static analysis check for `Reduction` as well? Should be very similar to `ScanLeft`

tests/static-analysis/scanleft.php
```diff
@@ -0,0 +1,41 @@
...
+use loophp\collection\Collection;
+use loophp\collection\Contract\Collection as CollectionInterface;
+
+$sum = static fn (int $carry, int $value): int => $carry + $value;
+$concat = static fn (?string $carry, string $string): string => sprintf('%s%s', (string) $carry, $string);
+
+/**
+ * @param CollectionInterface<int, int> $collection
+ */
+function scanleft_checkListInt(CollectionInterface $collection): void
```
AlexandruGG, line 21:
```suggestion
function scanLeft_checkListInt(CollectionInterface $collection): void
```

same with the filename

AlexandruGG: and in the other one

tests/static-analysis/pair.php
AlexandruGG:
```
╭─alex@AGMac ~/dev/collections/collection ‹feature/add-reduce-operation›
╰─$ php tests/static-analysis/pair.php
PHP Warning:  Undefined array key 1 in /Users/alex/dev/collections/collection/src/Operation/Pair.php on line 49

Warning: Undefined array key 1 in /Users/alex/dev/collections/collection/src/Operation/Pair.php on line 49
array(2) {
  [1]=>
  int(2)
  [3]=>
  NULL
}

File: src/Contract/Operation/ScanLeftable.php (hunk 6 above)
Comment (AlexandruGG) [reply src/Contract/Operation/ScanLeftable.php:27]: that's what I said as well, but why `V|T` and not `V`?

Comment (drupol) [reply src/Contract/Operation/ScanLeftable.php:27]: That said, there is another issue. I'm being fixing it.

File: src/Operation/Reduction.php (hunk 9 above)
Comment (drupol) [reply src/Operation/Reduction.php:31]: Done!

File: src/Contract/Operation/ScanRightable.php (hunk 8 above)
Comment (drupol) [reply src/Contract/Operation/ScanRightable.php]: Done!

File: src/Operation/ScanLeft.php (hunk 7 above)
Comment (drupol) [reply src/Operation/ScanLeft.php]: Thanks! I updated it anyway, I guess it's better now.

File: src/Contract/Operation/ScanLeftable.php (hunk 6 above)
Comment (drupol) [reply src/Contract/Operation/ScanLeftable.php]: Fixed.

File: tests/static-analysis/pair.php (hunk 5 above)
Comment (drupol) [reply tests/static-analysis/pair.php:29]: Weird, I don't have it locally. 

/me investigating.

File: tests/static-analysis/scanleft.php (hunk 10 above)
Comment (drupol) [reply tests/static-analysis/scanleft.php]: Done!

File: tests/static-analysis/pair.php (hunk 5 above)
Comment (drupol) [reply tests/static-analysis/pair.php:29]: Fixed in 995d26dc218d2bef41a54097a252d8c7c61ef93f !

**Comment by drupol** [comment]
I'm going to create separate PR's, it's too messy in here :) this PR was only supposed for adding the `Reduce` operation.

**Comment by AlexandruGG** [comment]
> I'm going to create separate PR's, it's too messy in here :) this PR was only supposed for adding the `Reduce` operation.

Yes please 🙌

**Comment by drupol** [comment]
> > I'm going to create separate PR's, it's too messy in here :) this PR was only supposed for adding the `Reduce` operation.
> 
> Yes please 🙌 

Everything's done already !

File: docs/pages/code/operations/reduce.php [hunk 11]
```diff
@@ -0,0 +1,22 @@
...
+
+include __DIR__ . '/../../../../vendor/autoload.php';
+
+$callback = static fn (int $carry, int $item): int => $carry + $item;
+
+$collection = Collection::empty()
+    ->reduce($callback); // []
+
+$collection = Collection::fromIterable(range(1, 5))
+    ->reduce($callback); // [15]
```
AlexandruGG, line 22:
```suggestion
    ->reduce($callback); // [4 => 15]
```

drupol:
Oops, looks like you added back all the commits I removed yesterday :S

AlexandruGG:
> Oops, looks like you added back all the commits I removed yesterday :S (see #153)

Yeah I just noticed, I had your branch already locally 😓

AlexandruGG:
> > Oops, looks like you added back all the commits I removed yesterday :S (see #153)
> 
> Yeah I just noticed, I had your branch already locally 😓

Sorry, can you fix it? There's only one commit I added (other than the master merge)

drupol:
Ok going to fix it.

drupol:
Job done.

To get it locally:

```
git fetch --all
git checkout feature/add-reduce-operation
git reset origin/feature/add-reduce-operation --hard
```

AlexandruGG (approved):
Looks great 🎉 

For the static analysis tests I always try to add some "failing" to ensure that the constraints we set work and make sense.

In this case the main constraint is that if the user doesn't provide an initial value then the default `null` will be used, which means the callback needs to be able to accept `null` as the _carry_ 😁

PR #145, opened 2021-07-14
refactor: Rename variables and properties.

drupol:
This PR:

* [x] Update variable names
* [x] Change the constructor type in `TypedIterator`
* [x] Break backward compatibility

### DIFF
--- a/spec/loophp/collection/Iterator/TypedIteratorSpec.php
+++ b/spec/loophp/collection/Iterator/TypedIteratorSpec.php
@@ -29,7 +29,7 @@ class TypedIteratorSpec extends ObjectBehavior
     {
         $data = [self::MAP_DATA, self::LIST_DATA];
 
-        $this->beConstructedWith($data);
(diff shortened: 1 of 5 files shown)


spec/loophp/collection/Iterator/TypedIteratorSpec.php
```diff
@@ -29,7 +29,7 @@ public function it_allows_array_of_any_type(): void
     {
         $data = [self::MAP_DATA, self::LIST_DATA];
 
-        $this->beConstructedWith($data);
+        $this->beConstructedWith(new ArrayIterator($data));
```
AlexandruGG, line 32: why do we not want to allow this to be constructed from `array`?

drupol: 1. We avoid mixing types.
2. I rather use the decorator pattern, it seems to be the way to go here.

PR #147, opened 2021-07-14
Show partition failure

AlexandruGG:
I noticed an issue trying to use the new `partition` method from **master**. I am not sure exactly what's causing it, but I wrote a test to showcase -> whenever we run `count` on one of the results from the partition operation it causes this fatal error:

`Fatal error: Generator passed to yield from was aborted without proper return and is unable to continue`

In the other project where I was trying to use this I dumped the collection before-after the `count`:

### Before
```PHP
loophp\collection\Collection^ {#5180
  -parameters: array:1 [
    0 => Generator {#5177
      executing: {
        ./vendor/loophp/collection/src/Operation/Filter.php:46 { …}
      }
      closed: false
    }
  ]
  -source: Closure(iterable $iterable): Iterator^ {#5148
    returnType: "Iterator"
    class: "loophp\collection\Collection"
  }
}
```

### After
```PHP
loophp\collection\Collection^ {#5180
  -parameters: array:1 [
    0 => Generator {#5177
      closed: true
    }
  ]
  -source: Closure(iterable $iterable): Iterator^ {#5148
    returnType: "Iterator"
    class: "loophp\collection\Collection"
  }
}
```

Based on this it looks like the generator appears closed

PR #148, opened 2021-07-15
Update Span and Partition operations.

drupol:
This PR:

* [x] Fix #147 
* [x] Fix Span operation which is having the same issue

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -22,11 +22,14 @@ fromCallable
 
 Create a collection from a callable.
 
-Signature: ``Collection::fromCallable(callable $callable, ...$parameters): Collection;``
+.. tip:: This can be very useful when working with a PHP `Generator`_, since it will allow the collection
+    object to behave as if the Generator was rewindable.
+
+Signature: ``Collection::fromCallable(callable $callable, iterable $parameters = []): Collection;``
 
 .. code-block:: php
 
-    $callback = static function () {
+    $callback = static function (): Generator {
         yield 'a';
         yield 'b';
         yield 'c';
@@ -49,6 +52,10 @@ fromIterable
 ~~~~~~~~~~~~
 
 Create a collection from an iterable.
+
+.. warning:: When instantiating from a PHP `Generator`_, the collection object will inherit its behaviour:
+    it will only be iterable a single time, and an exception will be thrown if multiple operations which attempt
+    to re-iterate are applied, for example ``count()``.
 
 Signature: ``Collection::fromIterable(iterable $iterable): Collection;``
 
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -669,7 +669,7 @@ final class Collection implements CollectionInterface
         return new self(
             Pipe::of()(
                 Partition::of()(...$callbacks),
-                Map::of()(static fn (Iterator $iterator): CollectionInterface => self::fromIterable($iterator))
+                Map::of()(static fn (array $callUserFuncParameters): CollectionInterface => self::fromCallable(...$callUserFuncParameters)),
             ),
             [$this->getIterator()]
         );
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -667,6 +667,7 @@ final class Collection implements CollectionInterface
 
     public function partition(callable ...$callbacks): CollectionInterface
     {
+        // TODO: Move this docblock above closure when https://github.com/phpstan/phpstan/issues/3770 lands.
         $mapCallback = static function (array $partitionResult): CollectionInterface {
             /**
              * @var Closure(Iterator<TKey, T>): Generator<TKey, T> $callback
@@ -787,6 +788,7 @@ final class Collection implements CollectionInterface
 
     public function span(callable ...$callbacks): CollectionInterface
     {
+        // TODO: Move this docblock above closure when https://github.com/phpstan/phpstan/issues/3770 lands.
         $mapCallback = static function (array $spanResult): CollectionInterface {
             /**
              * @var Closure(Iterator<TKey, T>): Generator<TKey, T> $callback
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -797,7 +797,7 @@ final class Collection implements CollectionInterface
             return self::fromCallable($callback, $parameters);
         };
 
-        return new self(Span::of()(Partition::of()(...$callbacks), Map::of()($mapCallback)), [$this->getIterator()]);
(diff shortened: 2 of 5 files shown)


src/Collection.php
```diff
@@ -154,14 +153,14 @@ final class Collection implements CollectionInterface
     private array $parameters;
 
     /**
-     * @var callable(mixed ...$parameters): iterable<TKey, T>
+     * @var callable(mixed ...$parameters): Iterator<TKey, T>
```
AlexandruGG, line 156: this change means we no longer want to allow this?

```PHP
Collection::fromCallable(static fn(): array => [1, 2, 3])
```

I don't see why this would not be allowed because `fromIterable` allows `array`, i.e.:
```PHP
Collection::fromIterable([1, 2, 3])
```

AlexandruGG: I guess we can argue that we want to keep it to only one type `Iterator`, but it feels like a restriction which should be justified in some way in the documentation

drupol: This is now no more valid, I got a better idea, I let you check it out ;)

drupol:
@AlexandruGG Let me know what you think. I think this is clearly an improvement.

AlexandruGG:
@drupol hey, so if I understand correctly the main issue is that `fromCallable` should be used whenever you want to instantiate the Collection from a Generator, right? And then it means that [this example](https://github.com/loophp/collection/pull/147#issuecomment-880437509) is kind of expected behaviour due to how PHP Generators work, right?

I'm thinking I can update the documentation to mention this and fix the static analysis as well

drupol:
Yes exactly, and it make sense actually, I don't know why I haven't though about this earlier.

AlexandruGG:
> Yes exactly, and it make sense actually, I don't know why I haven't though about this earlier.

Cool 😄. I thought about it at first but then I also was assuming that with Collection and ClosureIterator we want to still allow that. But it makes sense to keep it consistent with how a Generator would work by itself

AlexandruGG:
Leave this to me, I'll fix the static analysis and update the docs

drupol:
> > Yes exactly, and it make sense actually, I don't know why I haven't though about this earlier.
> 
> Cool 😄. I thought about it at first but then I also was assuming that with Collection and ClosureIterator we want to still allow that. But it makes sense to keep it consistent with how a Generator would work by itself

As a closure is the real immutable thing in PHP, we must use and abuse them 💪

drupol:
> Leave this to me, I'll fix the static analysis and update the docs

Thanks :) and ok, I don't touch this anymore.

drupol:
> @drupol hey, so if I understand correctly the main issue is that `fromCallable` should be used whenever you want to instantiate the Collection from a Generator, right? And then it means that [this example](https://github.com/loophp/collection/pull/147#issuecomment-880437509) is kind of expected behaviour due to how PHP Generators work, right?

Yes ! We could imagine fixing it by detecting if the `iterable` is a `Generator`, and then wrap it in a `CachingIterator`. We should tackle that in another PR anyway.

docs/pages/api.rst
```diff
@@ -22,11 +22,14 @@ fromCallable
 
 Create a collection from a callable.
 
-Signature: ``Collection::fromCallable(callable $callable, ...$parameters): Collection;``
+.. tip:: This can be very useful when working with a PHP `Generator`_, since it will allow the collection
+    object to behave as if the Generator was rewindable.
+
+Signature: ``Collection::fromCallable(callable $callable, iterable $parameters = []): Collection;``
```
AlexandruGG, line 28: don't forget to update these when changing method signatures 😬 
https://github.com/loophp/collection/pull/146

drupol: Oops! Thanks!

AlexandruGG (approved):
This should be good to go. Here's what I did:
- updated docs to mention the Generator behaviour. We can further update once #150 is ready
- fixed static analysis - required being a bit more verbose in the Collection.php methods and in the operations but works
- moved the new tests into the existing partition test

src/Collection.php
```diff
@@ -666,13 +667,17 @@ public function pair(): CollectionInterface
...
+            /**
+             * @var Closure(Iterator<TKey, T>): Generator<TKey, T> $callback
+             * @var array{0: Iterator<TKey, T>} $parameters
+             */
+            [$callback, $parameters] = $partitionResult;
+
+            return self::fromCallable($callback, $parameters);
+        };
+
+        return new self(Pipe::of()(Partition::of()(...$callbacks), Map::of()($mapCallback)), [$this->getIterator()]);
```
drupol, line 680: What do you think about doing this:

```diff
diff --git a/src/Collection.php b/src/Collection.php
index 640df98a..8042ecd5 100644
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -667,15 +667,11 @@ final class Collection implements CollectionInterface
 
     public function partition(callable ...$callbacks): CollectionInterface
     {
-        $mapCallback = static function (array $partitionResult): CollectionInterface {
+        $mapCallback =
             /**
-             * @var Closure(Iterator<TKey, T>): Generator<TKey, T> $callback
-             * @var array{0: Iterator<TKey, T>} $parameters
+             * @param array{0: Closure(Iterator<TKey, T>): Generator<TKey, T>, 1: array{0: Iterator<TKey, T>}} $partitionResult
              */
-            [$callback, $parameters] = $partitionResult;
-
-            return self::fromCallable($callback, $parameters);
-        };
+            static fn (array $partitionResult): CollectionInterface => self::fromCallable(...$partitionResult);
 
         return new self(Pipe::of()(Partition::of()(...$callbacks), Map::of()($mapCallback)), [$this->getIterator()]);
     }
```

drupol: So we don't `unwrap` and `wrap` parameters.

AlexandruGG: already tried that, doesn't work because PHPStan doesn't yet support reading doc blocks above closures: https://github.com/phpstan/phpstan/issues/3770

drupol: Oh... I'll add a TODO and merge!

AlexandruGG: in other cases it doesn't mind and just treats it as `callable` and `mixed`, but here it needs to know the return type for the `$callback` so that it can link it to the template types of `fromCallable`

drupol:
A nice one done! And a good lesson learned :)

drupol:
Thanks for the help @AlexandruGG !

AlexandruGG:
Thank you for fixing this 🚀

PR #151, opened 2021-07-17
Link to `stable` docs instead of `latest`

AlexandruGG:
This PR updates the links in the README to link to the `stable` version of the docs (last release), instead of `latest` (what's in `master`).


AlexandruGG:
@drupol I noticed that the docs get updated as soon as we merge into `master`, but I think that from the README we should link to the `stable` version (last release).

The user can always switch to latest if they want:
Screenshot 2021-07-17 at 13 03 51

What do you think?

drupol:
All good :)💪

drupol (approved):
:+1:

drupol:
Feel free to merge when you think it's ready!

PR #152, opened 2021-07-17
Add `Equals` Operation

AlexandruGG:
This PR:

* [x] Provides a new operation `Equals` which allows determining if two collections contain the same elements
* [x] Includes a small refactor for `IsEmpty` -> now returns a `Generator` so it can be piped. This is backwards-compatible at Collection level but a BC break at individual Operation level
* [x] Has documentation
* [x] Bonus: static analysis check for `diff`

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -708,6 +708,26 @@ Signature: ``Collection::duplicate(): Collection;``
             ->distinct()
             ->normalize() // [0 => 'a', 1 => 'c']
 
+equals
+~~~~~~
+
+Compare two collections for equality. Collections are considered *equal* if:
+
+* they have the same number of elements;
+* they contain the same elements, regardless of the order they appear in or their keys.
+
+Elements will be compared using strict equality (``===``).
+
+.. tip:: This operation enables comparing ``Collection`` objects in PHPUnit tests using
+    the dedicated `assertObjectEquals`_ assertion. 
+
+Interface: `Equalsable`_
+
+Signature: ``Collection::equals(Collection $other): bool;``
+
+.. literalinclude:: code/operations/equals.php
+  :language: php
+
 every
 ~~~~~
 
@@ -2433,6 +2453,7 @@ Signature: ``Collection::zip(iterable ...$iterables): Collection;``
 .. _DropWhileable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/DropWhileable.php
 .. _Dumpable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Dumpable.php
 .. _Duplicateable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Duplicateable.php
+.. _Equalsable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Equalsable.php
 .. _Everyable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Everyable.php
 .. _Explodeable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Explodeable.php
 .. _Falsyable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Falsyable.php
@@ -2520,6 +2541,7 @@ Signature: ``Collection::zip(iterable ...$iterables): Collection;``
 .. _Zipable: https://github.com/loophp/collection/blob/master/src/Contract/Operation/Zipable.php
 
 .. _array_flip(): https://php.net/array_flip
+.. _assertObjectEquals: https://phpunit.readthedocs.io/en/9.5/assertions.html#assertobjectequals
 .. _Countable: https://www.php.net/manual/en/class.countable.php
 .. _Criteria: https://www.doctrine-project.org/projects/doctrine-collections/en/1.6/index.html#matching
 .. _Doctrine Collections: https://github.com/doctrine/collections
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -176,7 +176,7 @@ or through the ``Collection`` object.
 When used separately, operations typically return a PHP `Generator`_ or an `Iterator`_.
 When used as a ``Collection`` method, operations fall into a few main categories based on the return type:
 
-1. Operations that return a ``boolean`` or ``scalar`` value: ``Contains``, ``Count``, ``Every``, ``Falsy``, ``Has``, ``IsEmpty``, ``Match`` (or ``MatchOne``), ``Nullsy``, ``Truthy``.
+1. Operations that return a ``boolean`` or ``scalar`` value: ``Contains``, ``Count``, ``Equals``, ``Every``, ``Falsy``, ``Has``, ``IsEmpty``, ``Match`` (or ``MatchOne``), ``Nullsy``, ``Truthy``.
 
 2. Operations that return a ``Collection`` of ``Collection`` objects: ``Partition``, ``Span``.
 
@@ -334,7 +334,7 @@ asyncMapN
 Asynchronously apply one or more supplied callbacks to every item of a collection and use the return value.
 
 .. tip:: This operation is best used when multiple callbacks need to be applied. If you only want to apply
-        a single callback, ``asyncMap`` should be prefered as it benefits from more specific type hints.
+        a single callback, ``asyncMap`` should be preferred as it benefits from more specific type hints.
 
 .. warning:: This method requires `amphp/parallel-functions <https://github.com/amphp/parallel-functions>`_ to be installed.
 
@@ -722,8 +722,9 @@ Elements will be compared using strict equality (``===``).
     the dedicated `assertObjectEquals`_ assertion. 
 
 .. warning:: Because this operation *needs to traverse both collections* to determine if
-    the same elements are contained within them, a performance cost is incurred. It is not
-    recommended to use this for potentially large collections.
+    the same elements are contained within them, a performance cost is incurred. Even though
+    the operation will stop as soon as it encounters an element of one collection that cannot
+    be found in the other, it is not recommended to use this for potentially large collections.
 
 Interface: `Equalsable`_
 
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -721,6 +721,10 @@ Elements will be compared using strict equality (``===``).
 .. tip:: This operation enables comparing ``Collection`` objects in PHPUnit tests using
     the dedicated `assertObjectEquals`_ assertion. 
 
+.. warning:: Because this operation *needs to traverse both collections* to determine if
+    the same elements are contained within them, a performance cost is incurred. It is not
+    recommended to use this for potentially large collections.
+
 Interface: `Equalsable`_
 
 Signature: ``Collection::equals(Collection $other): bool;``
--- a/src/Operation/Equals.php
+++ b/src/Operation/Equals.php
@@ -40,8 +40,22 @@ final class Equals extends AbstractOperation
                  *
                  * @return Generator<int, bool>
                  */
-                return static fn (Iterator $self): Generator => yield (iterator_count($other) === iterator_count($self))
-                        && yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($self);
+                return static function (Iterator $self) use ($other): Generator {
+                    $sameCount = true;
+
+                    do {
+                        if ($self->valid() !== $other->valid()) {
+                            $sameCount = false;
+
+                            break;
+                        }
+
+                        $self->next();
+                        $other->next();
+                    } while ($self->valid() || $other->valid());
+
+                    yield $sameCount && yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($self);
+                };
             };
     }
 }
--- a/src/Operation/Equals.php
+++ b/src/Operation/Equals.php
@@ -41,16 +41,18 @@ final class Equals extends AbstractOperation
                  * @return Generator<int, bool>
                  */
                 return static function (Iterator $iterator) use ($other): Generator {
+                    $sameCount = true;
+
                     while ($other->valid() && $iterator->valid()) {
                         $iterator->next();
                         $other->next();
                     }
 
                     if ($other->valid() !== $iterator->valid()) {
-                        yield false;
+                        $sameCount = false;
                     }
 
-                    yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($iterator);
+                    return yield $sameCount && yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($iterator);
                 };
             };
     }
--- a/src/Operation/Equals.php
+++ b/src/Operation/Equals.php
@@ -36,25 +36,21 @@ final class Equals extends AbstractOperation
              */
             static function (Iterator $other): Closure {
                 /**
-                 * @param Iterator<TKey, T> $self
+                 * @param Iterator<TKey, T> $iterator
                  *
                  * @return Generator<int, bool>
                  */
-                return static function (Iterator $self) use ($other): Generator {
-                    $sameCount = true;
-
-                    do {
-                        if ($self->valid() !== $other->valid()) {
-                            $sameCount = false;
-
-                            break;
-                        }
-
-                        $self->next();
+                return static function (Iterator $iterator) use ($other): Generator {
+                    while ($other->valid() && $iterator->valid()) {
+                        $iterator->next();
                         $other->next();
-                    } while ($self->valid() || $other->valid());
+                    }
 
-                    yield $sameCount && yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($self);
+                    if ($other->valid() !== $iterator->valid()) {
+                        yield false;
+                    }
+
+                    yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($iterator);
                 };
             };
     }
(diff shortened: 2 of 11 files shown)


src/Operation/Equals.php
```diff
@@ -0,0 +1,48 @@
...
+             */
+            static function (Iterator $other): Closure {
+                /** @var Closure(Iterator<TKey, T>): Generator<int, bool> $pipe */
+                $pipe = Pipe::of()(
+                    Diff::of()(...$other),
+                    IsEmpty::of(),
+                );
+
+                // Point free style.
+                return $pipe;
```
drupol, line 45: Be careful! [It's a trap!](https://i.kym-cdn.com/entries/icons/original/000/000/157/itsatrap.jpg) !

Using `Diff` means that you're going to go through the whole iterator.

But yet what you're trying to do is to compare if they are equal, in other words: find if there is **at least** two elements that are different.

It means that as soon as two elements are different, there is no need to go through the whole iterator, we know that they are different.

It's then a job for the `MatchOne` operation.

drupol (changes requested):

src/Operation/Equals.php
AlexandruGG: Hey @drupol, thanks for this; does this work if the elements are in different positions in the collections?

As in:

Collection (1, 2, 3)
Collection (2, 3, 1)

These should still be considered equal.

AlexandruGG: Or are you thinking we should consider them unequal?

I feel like there's scope to have 2 operations maybe: `equals` and `same`

drupol: > Hey @drupol, thanks for this; does this work if the elements are in different positions in the collections?
> 
> As in:
> 
> Collection (1, 2, 3)
> Collection (2, 3, 1)

No it doesn't do that sadly.



> I feel like there's scope to have 2 operations maybe: `equals` and `same`

Exact, we should have 2 different operations for that.

AlexandruGG: Cool, in that case:
- `equals` would need to check the entire collection, so this implementation should be fine
- `same` should stop as soon as an element is different or the end of one collection is reached, because we want them in the same positions

Regarding allowing the user to customise the comparison behaviour:
I want at least one of the operations (I'm thinking `equals`) to _only accept a single parameter_, otherwise it won't work with PHPUnit's `assertObjectEquals` method: https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/Object/ObjectEquals.php#L97-L102

Perhaps for `same` we can allow that with a signature: `Collection::same(Collection $other, ?callable $comparator = nul): bool`.

WDYT?

drupol: It seems ok to me, I guess we will refine all of these during the review.

src/Operation/Equals.php
```diff
@@ -0,0 +1,47 @@
...
+             * @return Closure(Iterator<TKey, T>): Generator<int, bool>
+             */
+            static function (Iterator $other): Closure {
+                /**
+                 * @param Iterator<TKey, T> $self
+                 *
+                 * @return Generator<int, bool>
+                 */
+                return static fn (Iterator $self): Generator => yield (iterator_count($other) === iterator_count($self))
+                        && yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($self);
```
AlexandruGG, line 44: @drupol I realised it's a bit trickier than I thought initially, can you think of how to do this better and still pass all the tests? (don't mind the current failures, it's just that some cannot be run on php 7.4 so I'll need to implement that)

drupol: It's the same issue as in the `Combine` operation. I'll give it a try later for sure!

drupol: I got the tests working by doing:

```diff
diff --git a/src/Operation/Equals.php b/src/Operation/Equals.php
index b758352a..4f71f4a4 100644
--- a/src/Operation/Equals.php
+++ b/src/Operation/Equals.php
@@ -40,8 +40,10 @@ final class Equals extends AbstractOperation
                  *
                  * @return Generator<int, bool>
                  */
-                return static fn (Iterator $self): Generator => yield (iterator_count($other) === iterator_count($self))
-                        && yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($self);
+                return static function (Iterator $self) use ($other): Generator {
+                    return yield (iterator_count($other) === iterator_count($self))
+                    && Pipe::of()(Diff::of()(...$other), IsEmpty::of())($self)->current();
+                };
             };
     }
 }
```

AlexandruGG: ooh good point, I didn't know about that one. I'll see if I can refactor a bit based on that

src/Operation/Equals.php
```diff
@@ -0,0 +1,61 @@
...
+                            $sameCount = false;
+
+                            break;
+                        }
+
+                        $self->next();
+                        $other->next();
+                    } while ($self->valid() || $other->valid());
+
+                    yield $sameCount && yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($self);
```
AlexandruGG, line 57: note we don't do the Pipe if `$sameCount` is _false_, so this more efficient than before, because as soon as we notice we've exhausted one of the iterators we know they cannot be equal

src/Operation/Equals.php
AlexandruGG: why is that one more efficient? We are still doing `iterator_count` on both of them. Let me know what you think of this version:

https://github.com/loophp/collection/pull/152/commits/11611a9170de8cc0d817d5a2ffec6c16736e4273

drupol: I didn't say that my fix was optimized.

When I need to do something like this, I always proceed by "iteration" (pun intended!), I start with a dumb/unoptimized version, and then I refine it multiple times.

Check how I would see the thing here:

```diff
diff --git a/src/Operation/Equals.php b/src/Operation/Equals.php
index 643719c8..222c43ed 100644
--- a/src/Operation/Equals.php
+++ b/src/Operation/Equals.php
@@ -40,21 +40,19 @@ final class Equals extends AbstractOperation
                  *
                  * @return Generator<int, bool>
                  */
-                return static function (Iterator $self) use ($other): Generator {
+                return static function (Iterator $iterator) use ($other): Generator {
                     $sameCount = true;
 
-                    do {
-                        if ($self->valid() !== $other->valid()) {
-                            $sameCount = false;
-
-                            break;
-                        }
-
-                        $self->next();
+                    while ($other->valid() && $iterator->valid()) {
+                        $iterator->next();
                         $other->next();
-                    } while ($self->valid() || $other->valid());
+                    }
+
+                    if ($other->valid() !== $iterator->valid()) {
+                        $sameCount = false;
+                    }
 
-                    yield $sameCount && yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($self);
+                    return yield $sameCount && yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($iterator);
                 };
             };
     }
```

I'm unsure about the last line.

Should we do: 

```php
return yield $sameCount && yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($iterator);
```

or 

```php
return yield $sameCount && Pipe::of()(Diff::of()(...$other), IsEmpty::of())($iterator)->current();
```

It would be very interesting to check if the second `yield` is triggered when `$sameCount` is `false`.

drupol: The only thing I don't like in this thing is the `Diff::of()(...$other)`, we really should find a better way, I have the impression that this is absolutely not the best option.

AlexandruGG: > I'm unsure about the last line.

I like the `yield from` version

drupol: It's not a matter of `yield` or `yield from`, I also prefer the `yield from`.

I'm afraid that those are as we have 2 yields, they are not evaluated as a real `Logical AND`.

Anyway, let me draft a quick example to prove my point.

drupol: And that's what I though: https://3v4l.org/t7Nqk

AlexandruGG: > Check how I would see the thing here:
Thank you for this suggestion, I simplified it even more and I think it looks good now

drupol: No, we cannot have 2 yields inside the function.

2 yields equals 2 returned value.

We must have only 1 yield.

AlexandruGG: > The only thing I don't like in this thing is the Diff::of()(...$other), we really should find a better way, I have the impression that this is absolutely not the best option.

I think you might have the right feeling there, but we can always improve on it later.

The reality is that with the way I envision `equals` working, we have to check all elements of both collections. I think it shouldn't be a problem as long as we mention in the documentation that there is a performance cost to it.

When we add `same` we will clarify that this is the more efficient one and the one to use if you want to check for `===` rather than just seeing if both collections contain the same elements.

drupol: Oh wait. I just noticed how you did it in `Collection` class.

drupol: Brilliant ! It's smart and beautiful, nice, I love it !

drupol: The only thing I would change:

```diff
diff --git a/src/Operation/Equals.php b/src/Operation/Equals.php
index d3b2b675..d0bae356 100644
--- a/src/Operation/Equals.php
+++ b/src/Operation/Equals.php
@@ -47,10 +47,10 @@ final class Equals extends AbstractOperation
                     }
 
                     if ($other->valid() !== $iterator->valid()) {
-                        yield false;
+                        return yield false;
                     }
 
-                    yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($iterator);
+                    return yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($iterator);
                 };
             };
     }
```

Like that, if the operation is used standalone, it won't yield 2 values, it will only yield 1 value.

AlexandruGG: > No, we cannot have 2 yields inside the function.
> 
> 2 yields equals 2 returned value.
> 
> We must have only 1 yield.

Ah you're right, I missed that 🤦‍♂️ . The operation wouldn't be properly usable by itself

drupol: No no, you should revert your [last commit](74b0e180f4736a90ecd932d4554310b94e1d3da2) and use [this diff here](https://github.com/loophp/collection/pull/152#discussion_r671887211).

AlexandruGG: > The only thing I would change:
> 
> ```diff
> diff --git a/src/Operation/Equals.php b/src/Operation/Equals.php
> index d3b2b675..d0bae356 100644
> --- a/src/Operation/Equals.php
> +++ b/src/Operation/Equals.php
> @@ -47,10 +47,10 @@ final class Equals extends AbstractOperation
>                      }
>  
>                      if ($other->valid() !== $iterator->valid()) {
> -                        yield false;
> +                        return yield false;
>                      }
>  
> -                    yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($iterator);
> +                    return yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($iterator);
>                  };
>              };
>      }
> ```
> 
> Like that, if the operation is used standalone, it won't (_potentially, when the first one is false_) yield 2 values, it will only yield 1 value.

Sadly you were right before and this also doesn't fix it:
https://3v4l.org/bJaaW#v8.0.8

However, the latest change fixes it. Based on:
https://3v4l.org/VMkTJ#v8.0.8

AlexandruGG: Ah the `return yield` 🎉

drupol: Yep, early returns are the way to go to avoid creating intermediary variables! https://3v4l.org/9rF6j#v8.0.8

AlexandruGG: excellent, thanks for the assistance on this

drupol:
Let's not forget to mention in the documentation that there is a "cost" with this operation.

AlexandruGG:
> Let's not forget to mention in the documentation that there is a "cost" with this operation.

Thanks! Didn't forget it, I was just debating whether to add it now or when we add the `same` operation so I can recommend using that one for better performance. I added a warning now anyway and we can modify it at that time

drupol:
By the way, I'm thinking about this optimization:

```diff
diff --git a/src/Operation/Equals.php b/src/Operation/Equals.php
index d0bae356..6f091110 100644
--- a/src/Operation/Equals.php
+++ b/src/Operation/Equals.php
@@ -50,7 +50,13 @@ final class Equals extends AbstractOperation
                         return yield false;
                     }
 
-                    return yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($iterator);
+                    foreach ($iterator as $current) {
+                        if (false === Contains::of()($current)($other)->current()) {
+                            return yield false;
+                        }
+                    }
+
+                    return yield true;
                 };
             };
     }
```

WDYT ?

docs/pages/api.rst
```diff
@@ -721,6 +721,10 @@ Elements will be compared using strict equality (``===``).
 .. tip:: This operation enables comparing ``Collection`` objects in PHPUnit tests using
     the dedicated `assertObjectEquals`_ assertion. 
 
+.. warning:: Because this operation *needs to traverse both collections* to determine if
+    the same elements are contained within them, a performance cost is incurred. It is not
+    recommended to use this for potentially large collections.
```
drupol, line 726: :100: !

drupol:
We can even use `Every` for that! Which is basically the same !

```diff
diff --git a/src/Operation/Equals.php b/src/Operation/Equals.php
index d0bae356..52304548 100644
--- a/src/Operation/Equals.php
+++ b/src/Operation/Equals.php
@@ -50,7 +50,12 @@ final class Equals extends AbstractOperation
                         return yield false;
                     }
 
-                    return yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($iterator);
+                    return yield from Pipe::of()(
+                        Every::of()
+                            (static fn(): bool => false)
+                            (static fn($current): bool => Contains::of()($current)($other)->current()),
+                        Head::of()
+                    )($iterator);
                 };
             };
     }
```

AlexandruGG:
> By the way, I'm thinking about this optimization:
> 
> ```diff
> diff --git a/src/Operation/Equals.php b/src/Operation/Equals.php
> index d0bae356..6f091110 100644
> --- a/src/Operation/Equals.php
> +++ b/src/Operation/Equals.php
> @@ -50,7 +50,13 @@ final class Equals extends AbstractOperation
>                          return yield false;
>                      }
>  
> -                    return yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($iterator);
> +                    foreach ($iterator as $current) {
> +                        if (false === Contains::of()($current)($other)->current()) {
> +                            return yield false;
> +                        }
> +                    }
> +
> +                    return yield true;
>                  };
>              };
>      }
> ```
> 
> I think this would be better. Indeed, the `Diff` operation will traverse the whole collection to compute a diff. A diff can contain 1, 2, 3... n values that are different. In our case, we only need to check if there's only one which is different, then return false.
> 
> WDYT ?

Hmm interesting one! `Diff` is basically doing a `Filter`, but we want to stop early. Could we use `TakeWhile` for that somehow?

drupol:
> Hmm interesting one! `Diff` is basically doing a `Filter`, but we want to stop early. Could we use `TakeWhile` for that somehow?

This is what the `Match` and `Every` operations are doing, they are not using a `TakeWhile` but a `DropWhile` instead.

AlexandruGG:
> > Hmm interesting one! `Diff` is basically doing a `Filter`, but we want to stop early. Could we use `TakeWhile` for that somehow?
> 
> This is what the `Match` and `Every` operations are doing, they are not using a `TakeWhile` but a `DropWhile` instead.

I just noticed this after I wrote it 😄

drupol:
This is what I would propose:

```diff
diff --git a/src/Operation/Equals.php b/src/Operation/Equals.php
index d0bae356..dcbfe2d0 100644
--- a/src/Operation/Equals.php
+++ b/src/Operation/Equals.php
@@ -50,7 +50,10 @@ final class Equals extends AbstractOperation
                         return yield false;
                     }
 
-                    return yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($iterator);
+                    return yield from Every::of()
+                        (static fn(): bool => false)
+                        (static fn($current): bool => Contains::of()($current)($other)->current())
+                        ($iterator);
                 };
             };
     }
```

drupol:
Good for me now! Feel free to merge whenever you want!

AlexandruGG:
> This is what I would propose:
> 
> ```diff
> diff --git a/src/Operation/Equals.php b/src/Operation/Equals.php
> index d0bae356..dcbfe2d0 100644
> --- a/src/Operation/Equals.php
> +++ b/src/Operation/Equals.php
> @@ -50,7 +50,10 @@ final class Equals extends AbstractOperation
>                          return yield false;
>                      }
>  
> -                    return yield from Pipe::of()(Diff::of()(...$other), IsEmpty::of())($iterator);
> +                    return yield from Every::of()
> +                        (static fn(): bool => false)
> +                        (static fn($current): bool => Contains::of()($current)($other)->current())
> +                        ($iterator);
>                  };
>              };
>      }
> ```
> 
> And like that, we gracefully get rid of the `Diff::of()(...$other)` !

This is great, I used it just had to tweak it a bit to satisfy phpcs and then psalm 😄

drupol (approved):
:+1:

AlexandruGG:
> Good for me now! Feel free to merge whenever you want!

I'll call it a day and give this a fresh final look tomorrow! But feel free to approve and I will merge it if I'm not changing anything.

AlexandruGG:
@drupol thanks for all the help on this! 💯

PR #153, opened 2021-07-18
Improve static analysis tests

drupol:
This PR:

This PR depends on #161


AlexandruGG:
👀

src/Contract/Operation/Reductionable.php
```diff
@@ -18,11 +19,16 @@
...
      *
-     * @param mixed $initial
+     * @template V
+     * @template W
      *
-     * @return Collection<TKey, T>
+     * @param callable(V|W, T, TKey, Iterator<TKey, T>): W $callback
+     * @param V $initial
+     *
+     * @return Collection<TKey, V|W>
```
AlexandruGG, line 31: shouldn't these be the same as `Reduceable`? Instead here we have 2 parameter types

https://github.com/loophp/collection/blob/master/src/Contract/Operation/Reduceable.php#L24-L29

AlexandruGG:
There has been [a phpstan release today](https://github.com/phpstan/phpstan/releases) which is causing some new errors, I'll see if I can sort them out

AlexandruGG:
> There has been [a phpstan release today](https://github.com/phpstan/phpstan/releases) which is causing some new errors, I'll see if I can sort them out

Psalm and PHPStan are having some disagreements now on this whole non-empty-string thing 🙄. I bet there's going to be a fix/tweak

AlexandruGG:
could you use this commit https://github.com/loophp/collection/pull/153/commits/ff9c4b6ba9c44cc9e2f1c1dc4ac22fd7b878f25c as an example to add some "failure cases" for the other operations? If we are not "stress-testing" the type annotations I'm not confident that they are correct or that the static analysis tests are adding any value

src/Contract/Operation/Reductionable.php
drupol: Indeed, you're right.

drupol: Fixed in b44b87f689da7e7f9a0ac205302a4b697347bb22

drupol:
@AlexandruGG I fixed the issue with `Reduction`, however, regarding the `non-empty-string` with phpstan I have some issues.
When I fix it, psalm complains... and when I don't phpstan complains!

Do you think we should fix the version of PHPStan until this is fixed upstream?

AlexandruGG:
> @AlexandruGG I fixed the issue with `Reduction`, however, regarding the `non-empty-string` with phpstan I have some issues.
> When I fix it, psalm complains... and when I don't phpstan complains!
> 
> Do you think we should fix the version of PHPStan until this is fixed upstream?

Yeah I had that issue as well, yeah let's fix the version for a bit and see what happens

drupol:
@AlexandruGG I fixed the PHPStan issues, now Psalm complains.

Which version of PHPStan do you advice to use?

src/Contract/Operation/Reductionable.php
```diff
@@ -22,9 +23,13 @@ interface Reductionable
      *
      * @see https://loophp-collection.readthedocs.io/en/stable/pages/api.html#reduction
      *
-     * @param mixed $initial
+     * @template V
+     * @template W
```
AlexandruGG, line 27: I don't understand why this is different from [Reduceable](https://github.com/loophp/collection/blob/master/src/Contract/Operation/Reduceable.php#L26).

Same of ScanLeftable and ScanRightable, shouldn't they all be the same?

drupol: You're right, there was an inconsistency, I just fixed it.

However, `ScanLeft` and `ScanRight` might return the initial value if the Iterator is empty, this is why it might return `V` or `W`.

AlexandruGG (approved):

drupol:
Best news of the day! Thanks for the review!

PR #154, opened 2021-07-18
Fix: Prevent warning when using odd amount of items.

drupol:
This PR:

* [x] Fix warnings when using the Pair operation and when having an odd number of items in the collection.


AlexandruGG (approved):
Nice 😄

PR #155, opened 2021-07-19
Add Operation: `Same`

AlexandruGG:
This PR:

* [x] Provides a new operation `Same` which allows determining if two collections are the same
* [x] Allows custom item comparison via a callable -> if this is not provided, by default we consider two collections the same if the have the same items and keys, in the same order
* [x] Has documentation

Follows #152.

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -716,15 +716,16 @@ Compare two collections for equality. Collections are considered *equal* if:
 * they have the same number of elements;
 * they contain the same elements, regardless of the order they appear in or their keys.
 
-Elements will be compared using strict equality (``===``).
+Elements will be compared using strict equality (``===``). If you want to customize how elements
+are compared or the order in which the keys/values appear is important, use the ``same`` operation.
 
 .. tip:: This operation enables comparing ``Collection`` objects in PHPUnit tests using
     the dedicated `assertObjectEquals`_ assertion. 
 
 .. warning:: Because this operation *needs to traverse both collections* to determine if
-    the same elements are contained within them, a performance cost is incurred. Even though
-    the operation will stop as soon as it encounters an element of one collection that cannot
-    be found in the other, it is not recommended to use this for potentially large collections.
+    the same elements are contained within them, a performance cost is incurred. The operation will stop
+    as soon as it encounters an element of one collection that cannot be found in the other. However, 
+    it is not recommended to use it for potentially large collections, where ``same`` can be used instead.
 
 Interface: `Equalsable`_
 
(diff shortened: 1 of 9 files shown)


AlexandruGG:
@drupol hey, this PR still needs some work but let me know if you're happy with the implementation itself. Very similar to Equals, with a few minor differences. It's actually simpler.

AlexandruGG:
Also, should we just ignore this phpcs rule for the whole project, or at least `src/Operation`? 😄

drupol:
About PHPCS: Let's add the rule in the file header for now.

Regarding the algorithm used in here, I think we could get rid of local variables using early returns.

AlexandruGG:
> About PHPCS: Let's add the rule in the file header for now.

Yep that's what I will do for this PR, but if we know for sure we will have this issue in many Operations due to currying and type annotations, would it not be good to disable it at least for that namespace? I think it makes the code look a bit more "messy" with the ignores in each file

AlexandruGG:
> Regarding the algorithm used in here, I think we could get rid of local variables using early returns.

Yep that's what I had initially but ran into some php-cs-fixer issues, have a look at my comments here: https://gist.github.com/AlexandruGG/6b201e47bc4e332f72d7080132cc944a

AlexandruGG:
> > Regarding the algorithm used in here, I think we could get rid of local variables using early returns.
> 
> Yep that's what I had initially but ran into some php-cs-fixer issues, have a look at my comments here: https://gist.github.com/AlexandruGG/6b201e47bc4e332f72d7080132cc944a

I updated the PR with an alternative. It's a bit more verbose but it works if we want to avoid the local variable

drupol:
@AlexandruGG I pushed a minor commit so we don't forget to fix later on!

drupol (dismissed):
I will be busy today and I don't know if I will be available that much, but the PR is fine for me, feel free to merge it when the tests are done and you feel like it's good.

AlexandruGG:
> @AlexandruGG I pushed a minor commit so we don't forget to fix later on!

@drupol thank you! Weirdly enough, I couldn't manage to reproduce that issue in a separate project, no matter how hard I tried 😄

AlexandruGG:
> I will be busy today and I don't know if I will be available that much, but the PR is fine for me, feel free to merge it when the tests are done and you feel like it's good.

@drupol thank you! Sorry but the last commit dismissed your review and GitHub won't let me merge it without the approval 😄

drupol (approved):
Quickly passing to approve, thanks for your work :) if you have time to check #153 as well it could be nice.

AlexandruGG:
> Quickly passing to approve, thanks for your work :) if you have time to check #153 as well it could be nice.

Thank you! Sure I'll have a look at that one this evening 😁

PR #156, opened 2021-07-20
Add API Docs Description to Interface Methods

AlexandruGG:
This PR:

* [x] Adds descriptions to Collection interface methods based on the API docs
* [x] Fixed/tweaked a few things in the API docs 

Related to https://github.com/loophp/collection/discussions/135#discussioncomment-1023285.

### DIFF
--- a/src/Contract/Operation/Dropable.php
+++ b/src/Contract/Operation/Dropable.php
@@ -18,7 +18,9 @@ use loophp\collection\Contract\Collection;
 interface Dropable
 {
     /**
-     * Skip the n items of a collection.
+     * Drop the `n` first items of the collection.
+     *
+     * @see https://loophp-collection.readthedocs.io/en/stable/pages/api.html#drop
      *
      * @param int ...$counts
      *
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -1317,6 +1317,8 @@ last
 
 Extract the last element of a collection, which must be finite and non-empty.
 
+The ``current`` operation can then be used to extract the item out of the collection.
+
 Interface: `Lastable`_
 
 Signature: ``Collection::last(): Collection;``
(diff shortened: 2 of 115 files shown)


src/Contract/Operation/Dropable.php
```diff
@@ -18,7 +18,9 @@
 interface Dropable
 {
     /**
-     * Skip the n items of a collection.
+     * Drop the `n` first items of the collection.
+     *
+     * @see https://loophp-collection.readthedocs.io/en/stable/pages/api.html#drop
```
AlexandruGG, line 23: @drupol I found this operation quite weird, especially because it takes variadic parameters and then adds them up.

```PHP
Collection::fromIterable([1, 2, 3, 4, 5])->drop(1, 4); 
// result -> []
// equivalent to:
Collection::fromIterable([1, 2, 3, 4, 5])->drop(5); 
```

In this example it drops the first 1+4=5 elements of the collection. As a user I wouldn't expect this, maybe I would expect it to drop keys 1 and 4.

Is there a use case for variadic parameters here? I would propose to remove that and only allow a single count, which is the number of items to drop.

drupol: I agree with that point, this would be a BC break, but I'm definitely ok with this in v5.

AlexandruGG:
The static analysis failures are all due to the `non-empty-string` PHPStan issues.

AlexandruGG:
@drupol this is ready for review. The documentation was good, but now it's looking even better!

drupol (approved):
Excellent, thanks !!!! :100:

AlexandruGG:
@drupol thanks for taking a look at this! 

I can't actually merge it because of the failing required checks (due to the PHPStan issue - https://github.com/phpstan/phpstan/issues/5372).

Could you merge it for me please?

drupol:
Amazing work, thanks again :)

Issue #118, opened 2021-06-29
Typed properties


jaroslavlibal:
Hello,

I have a question related to typed collections.

All classes in this library are final and we are expected to use the [composition pattern](https://loophp-collection.readthedocs.io/en/latest/pages/usage.html#extending-collection).

However I would like to use typed collections eg. PersonCollection as the collection of Person entities, including typehinting in business methods.

PHP Stan / PSALM generics are not enough for my use case as I would like to add some specific methods eg. AddressCollection getter to (and only to) the PersonCollection.

The implementation of my use-case seems more complicated when using composition as opposed to some AbstractCollection extension as I will have to implement all the basic CollectionInterface methods including `count()` or `getIterator()`.

I understand that this is a library concept, I just want to make sure I understand it correctly.

Or is there some possibility that hasn't occurred to me? Of course I can think of e.g. creating the mentioned AbstractCollection which will no longer be final, but I don't know if that's a conceptual solution.

Thanks!

drupol:
Hi !

When it comes to final keywords, there are a lot of post ([#1](https://ocramius.github.io/blog/when-to-declare-classes-final/), [#2](https://matthiasnoback.nl/2018/09/final-classes-by-default-why/)) discussion about that, and I have the feeling that you've understood the concept.

Basically I do not provide an abstracted version of the Collection because I want to ship a Collection class that works well in any circumstances, with a well defined scope and business logic.

If you want, you are free to implement your own Abstract collection class (example: https://gist.github.com/drupol/c73c7b092288833d2546edaac213bea5) and start from there for sure. On that example, all the methods are implemented, it would be possible to have a reduced version of it by specifying each interface that you want to have as well instead.

Let me know if this is helping you.

jaroslavlibal:
Hello,

thank you very much for your comprehensive reply!

I understand the reasons why it is implemented this way and will implement my own AbstractCollection class.

Thank you for your time and I hope this issue will help others who may ask a similar question in the future.

PR #158, opened 2021-07-24
Introduce helpers for static constructors

AlexandruGG:
This PR provides helper functions that can be used to instantiate Collection.
The main advantages are:
- fewer characters to type
- less import aliasing, e.g. when using the [CollectionInterface](https://github.com/loophp/collection/blob/master/src/Contract/Collection.php) as param or return type users always need to have 2 imports with the existing static constructors - one for the interface and one for the [Collection](https://github.com/loophp/collection/blob/master/src/Collection.php) class (see discussion link below)

* [x] Provides helper functions
* [ ] Has unit tests (phpspec)
* [ ] Has documentation

### DIFF
--- a/spec/loophp/collection/CollectionSpec.php
+++ b/spec/loophp/collection/CollectionSpec.php
@@ -29,6 +29,7 @@ use PhpSpec\Exception\Example\MatcherException;
 use PhpSpec\ObjectBehavior;
 use stdClass;
 use function gettype;
+use function loophp\collection\Utils\collectIterable;
 use const INF;
 use const PHP_EOL;
 use const PHP_VERSION_ID;
@@ -36,6 +37,13 @@ use const PHP_VERSION_ID;
 class CollectionSpec extends ObjectBehavior
 {
     private const PHP_8 = 80_000;
+
+    public function getMatchers(): array
+    {
+        return [
+            'beSameAs' => static fn (CollectionInterface $subject, CollectionInterface $expected): bool => $subject->same($expected),
+        ];
+    }
 
     public function it_can_all(): void
     {
@@ -436,6 +444,13 @@ class CollectionSpec extends ObjectBehavior
             ->beConstructedThrough('fromIterable', [new ArrayObject([1, 2, 3])]);
 
         $this->shouldImplement(Collection::class);
+    }
+
+    public function it_can_be_constructed_with_collectIterable(): void
+    {
+        $input = [1, 2, 3];
+
+        $this::fromIterable($input)->shouldBeSameAs(collectIterable($input));
     }
 
     public function it_can_be_instantiated_with_withClosure(): void
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,8 @@
     "autoload": {
         "psr-4": {
             "loophp\\collection\\": "./src/"
-        }
+        },
+        "files": ["src/Utils/helpers.php"]
     },
     "autoload-dev": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,9 @@
         "psr-4": {
             "loophp\\collection\\": "./src/"
         },
-        "files": ["src/Utils/helpers.php"]
+        "files": [
+            "src/Utils/helpers.php"
+        ]
     },
     "autoload-dev": {
         "psr-4": {
--- a/src/Utils/helpers.php
+++ b/src/Utils/helpers.php
@@ -1,0 +1,31 @@
+<?php
+
+/**
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace loophp\collection\Utils;
+
+use loophp\collection\Collection;
+use loophp\collection\Contract\Collection as CollectionInterface;
+use function function_exists;
+
+if (!function_exists('collectIterable')) {
+    /**
+     * @pure
+     *
+     * @template TKey
+     * @template T
+     *
+     * @param iterable<TKey, T> $iterable
+     *
+     * @return CollectionInterface<TKey, T>
+     */
+    function collectIterable(iterable $iterable): CollectionInterface
+    {
+        return Collection::fromIterable($iterable);
+    }
+}
--- a/src/Utils/helpers.php
+++ b/src/Utils/helpers.php
@@ -11,6 +11,7 @@ namespace loophp\collection\Utils;
 
 use loophp\collection\Collection;
 use loophp\collection\Contract\Collection as CollectionInterface;
+
 use function function_exists;
 
 if (!function_exists('collectIterable')) {


spec/loophp/collection/CollectionSpec.php
```diff
@@ -37,6 +38,13 @@ class CollectionSpec extends ObjectBehavior
 {
     private const PHP_8 = 80_000;
 
+    public function getMatchers(): array
+    {
+        return [
+            'beSameAs' => static fn (CollectionInterface $subject, CollectionInterface $expected): bool => $subject->same($expected),
+        ];
+    }
```
AlexandruGG, line 46: adding a custom matcher based on: https://www.phpspec.net/en/stable/cookbook/matchers.html#inline-matcher

Couldn't find a better way to test the function, let me know if you can think of any

AlexandruGG:
@drupol does this look good to you? If so, I can add functions for the other constructors

spec/loophp/collection/CollectionSpec.php
drupol: This is a good way indeed! And I don't see any other option at the moment.

drupol:
Hi Alex,

Static constructors are already helpers and I don't like adding this so much.

Even after reading the description, I'm still having troubles seeing the advantages of this.

The documentation is hopefully written correctly now to explain how to create Collection, why would we want to have a function that does it on top of what's already existing?

AlexandruGG:
> Hi Alex,
> 
> Static constructors are already helpers and I don't like adding this so much.
> 
> Even after reading the description, I'm still having troubles seeing the advantages of this.
> 
> The documentation is hopefully written correctly now to explain how to create Collection, why would we want to have a function that does it on top of what's already existing?

Hey @drupol,

Thanks for the feedback. I had a few conflicting thoughts about this as well, but I still think I would find it useful.

I think it would be great if we could list both the **pros** and **cons** of this. These are the ones I can think of:

**Pros**:
- no need for two separate `CollectionInterface` and `Collection` imports (more details in the [discussion](https://github.com/loophp/collection/discussions/135#discussioncomment-1016598))
- less characters to write. Might not seem like a big one, but if you you're writing many `Collection::fromIterable` vs `collectIterable` it can make a difference

**Cons**:
- "duplicating" functionality of the existing static constructors -> since the functions would be calling the static constructors anyway, the code is not really duplicated 
- global functions might clash with other user-defined functions -> since we use `function_exists`, in cases where `collectIterable` already exists it just wouldn't be defined. Those users would have to choose if they want to change their own function or just use the static constructors
- using global functions in general -> these functions are [namespaced](https://github.com/loophp/collection/pull/158/files#diff-41d990dda6814e072d7e35aef8822db3a05cb819602dca63c1ff8d81902060cdR10) so an import would always be needed

We could argue that the **pros** are not that significant and I partly agree, but would be nice to hear your reasoning.

drupol:
> **Pros**:
> 
>     * no need for two separate `CollectionInterface` and `Collection` imports (more details in the [discussion](https://github.com/loophp/collection/discussions/135#discussioncomment-1016598))

But why would this be a pro? I don't see any issue of having 1 or 2 `use` statements in the header of the file. I never find it problematic.

>     * less characters to write. Might not seem like a big one, but if you you're writing many `Collection::fromIterable` vs `collectIterable` it can make a difference

Yeah... but we have autocompletion and I never write it completely, but rely on my IDE to autocomplete it. So here, I don't see the difference between the 2 options.

> **Cons**:
> 
>     * "duplicating" functionality of the existing static constructors -> since the functions would be calling the static constructors anyway, the code is not really duplicated

Yeah. I also made it clear since I started to write this library: avoid repeated code and let the user deal with userland implementations.
This is the reason why I do not provide an `abstract Collection class`, because the use can implement it himself (see #118).

>     * global functions might clash with other user-defined functions -> since we use `function_exists`, in cases where `collectIterable` already exists it just wouldn't be defined. Those users would have to choose if they want to change their own function or just use the static constructors

True. I could have use functions for creating all the Collection operations, but I choose not to and created Classes. So for me creating functions for constructors doesn't make sense. This is the reason why we have multiple static constructors in Collection.

>     * using global functions in general -> these functions are [namespaced](https://github.com/loophp/collection/pull/158/files#diff-41d990dda6814e072d7e35aef8822db3a05cb819602dca63c1ff8d81902060cdR10) so an import would always be needed

Exactly, so, 1 or 2 `use` statements, I don't really see the point of doing that.

> We could argue that the **pros** are not that significant and I partly agree, but would be nice to hear your reasoning.

I added some feedback below each argument.

I usually like when people contributing inject their ideas, but honestly, for this one I don't follow you :disappointed:.
I think Collection should stay simple and clean. I'm already upset against me because of the `Utils/CallbacksArrayReducer.php` which shouldn't be there, so, adding helpers... no.

Maybe we should think about all these things in an extra-package, something like `loophp/collection-utils` where we could provide those things and have more flexibility, user-friendlyness and even more (an abstract Collection class?) ?

AlexandruGG:
> Hello there,
> 
> > **Pros**:
> > ```
> > * no need for two separate `CollectionInterface` and `Collection` imports (more details in the [discussion](https://github.com/loophp/collection/discussions/135#discussioncomment-1016598))
> > ```
> 
> But why would this be a pro? I don't see any issue of having 1 or 2 `use` statements in the header of the file. I never find it problematic.
> 
> > ```
> > * less characters to write. Might not seem like a big one, but if you you're writing many `Collection::fromIterable` vs `collectIterable` it can make a difference
> > ```
> 
> Yeah... but we have autocompletion and I never write it completely, but rely on my IDE to autocomplete it. So here, I don't see the difference between the 2 options.
> 
> > **Cons**:
> > ```
> > * "duplicating" functionality of the existing static constructors -> since the functions would be calling the static constructors anyway, the code is not really duplicated
> > ```
> 
> Yeah. I also made it clear since I started to write this library: avoid repeated code and let the user deal with userland implementations.
> This is the reason why I do not provide an `abstract Collection class`, because the use can implement it himself (see #118).
> 
> > ```
> > * global functions might clash with other user-defined functions -> since we use `function_exists`, in cases where `collectIterable` already exists it just wouldn't be defined. Those users would have to choose if they want to change their own function or just use the static constructors
> > ```
> 
> True. I could have use functions for creating all the Collection operations, but I choose not to and created Classes. So for me creating functions for constructors doesn't make sense. This is the reason why we have multiple static constructors in Collection.
> 
> > ```
> > * using global functions in general -> these functions are [namespaced](https://github.com/loophp/collection/pull/158/files#diff-41d990dda6814e072d7e35aef8822db3a05cb819602dca63c1ff8d81902060cdR10) so an import would always be needed
> > ```
> 
> Exactly, so, 1 or 2 `use` statements, I don't really see the point of doing that.
> 
> > We could argue that the **pros** are not that significant and I partly agree, but would be nice to hear your reasoning.
> 
> I added some feedback below each argument.
> 
> I usually like when people contributing inject their ideas, but honestly, for this one I don't follow you 😞.
> I think Collection should stay simple and clean. I'm already upset against me because of the `Utils/CallbacksArrayReducer.php` which shouldn't be there, so, adding helpers... no.
> 
> Maybe we should think about all these things in an extra-package, something like `loophp/collection-utils` where we could provide those things and have more flexibility, user-friendlyness and even more (an abstract Collection class?) ?

Fair enough, I think the pros are not that great for this one. Thanks for taking the time to go through this!

drupol:
Thanks for the constructive discussion Alex!

PR #165, opened 2021-07-30
SA: Update operations.

drupol:
This PR:

* [x] Fix some typing informations
* [x] Investigated on why Psalm >= 4.9.0 trigger issues with `scanLeft` and `scanRight` SA tests. (After bisecting Psalm, the commit introducing the issue is https://github.com/vimeo/psalm/commit/acc7ee261c106b4ae91f333ff4a9939b6e858bf3 from PR https://github.com/vimeo/psalm/pull/6072). Turns out that it was normal actually. To be confirmed.

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -702,15 +702,19 @@ final class Collection implements CollectionInterface
     public function partition(callable ...$callbacks): CollectionInterface
     {
         // TODO: Move this docblock above closure when https://github.com/phpstan/phpstan/issues/3770 lands.
-        $mapCallback = static function (array $partitionResult): CollectionInterface {
+        $mapCallback =
             /**
-             * @var Closure(Iterator<TKey, T>): Generator<TKey, T> $callback
-             * @var array{0: Iterator<TKey, T>} $parameters
+             * @param array{0: (Closure(Iterator<TKey, T>): Generator<TKey, T>), 1: (array{0: Iterator<TKey, T>})} $partitionResult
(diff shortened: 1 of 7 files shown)


AlexandruGG:
@drupol hey, indeed this looks like an issue with Psalm. Could you reproduce it at https://psalm.dev/ so we can also compare with PHPStan and open an issue?

drupol:
> @drupol hey, indeed this looks like an issue with Psalm. Could you reproduce it at https://psalm.dev/ so we can also compare with PHPStan and open an issue?

No, I haven't had the time to do it yet, I will most probably focus on this in the evening.

drupol:
@AlexandruGG I think it's a false positive. Actually Psalm was right, the `$concat` callback has a nullable `$carry` parameter, I didn't expected that.

Could you please let me know if it's ok for you?

AlexandruGG:
> @AlexandruGG I think it's a false positive. Actually Psalm was right, the `$concat` callback has a nullable `$carry` parameter, I didn't expected that.
> 
> Could you please let me know if it's ok for you?

Yep you're right!

Comparing to `Reduce`, the scans will always add the `initial` parameter passed in, which has a default `null`. Which means if the Collection is empty then you will be left with a `Collection<int, null>`. Whereas with `Reduce` you always get the type returned by the callback, and if the collection is empty it will continue to be empty

AlexandruGG (approved):

PR #166, opened 2021-07-31
Fix Column Typing + SA Check

AlexandruGG:
This PR:

* [x] Fixes the type hints for `Column`/`Columnable` -> this operation supports any "column" key type as parameter, not just `int|string`. Furthermore, the return type has to be flexible because it will not be `T`, but rather whatever type was associated with that key, so it has to be `mixed`
* [x] Has documentation

Note: I spotted this issue while trying to do this in another project:

```PHP
/** @return Collection<int, int> */
public function getIds(): Collection 
{
  // ...
  /** @var list<array{id: numeric-string}> $results */
  $results = $qb->getQuery()->getResult();

  return Collection::fromIterable($results)
    ->column('id')
    ->map(static fn (string $id): (int) $id);  
            // -> I got an error here because the map callable was expecting type `array{id: numeric-string}`
}

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -463,6 +463,9 @@ column
 
 Return the values from a single column in the input iterables.
 
+.. tip:: If the iterables you are selecting from are ``Generators``, the operation will allow
+         accessing keys of any type, not just ``int|string``.
+
 Interface: `Columnable`_
 
 Signature: ``Collection::column($column): Collection;``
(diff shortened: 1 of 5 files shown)


drupol:
This depends on #166

drupol:
All good for me here, merge anytime ! :)

AlexandruGG:
> All good for me here, merge anytime ! :)

Turned on auto-merge 😁

drupol (approved):
:+1:

PR #167, opened 2021-07-31
refactor: Update `Reject` operation

drupol:
This PR:

* [x]  Use Filter based algorithm
* [x]   Use point free style.

### DIFF
--- a/src/Operation/Reject.php
+++ b/src/Operation/Reject.php
@@ -37,30 +37,24 @@ final class Reject extends AbstractOperation
              *
              * @return Closure(Iterator<TKey, T>): Generator<TKey, T>
              */
-            static fn (callable ...$callbacks): Closure =>
-                /**
-                 * @param Iterator<TKey, T> $iterator
-                 *
-                 * @return Generator<TKey, T>
-                 */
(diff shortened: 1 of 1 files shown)


src/Operation/Reject.php
```diff
@@ -37,30 +37,24 @@ public function __invoke(): Closure
...
-
-                    foreach ($iterator as $key => $current) {
-                        $result = CallbacksArrayReducer::or()($callbacks, $current, $key, $iterator);
-
-                        if (false === $result) {
-                            yield $key => $current;
-                        }
-                    }
-                };
+            static function (callable ...$callbacks): Closure {
```
AlexandruGG, line 40: I can't tell for sure but does this closure have doc block type hints for the callables?

AlexandruGG (approved):

AlexandruGG:
Oops didn't realise this was set for auto-merge 😄

src/Operation/Reject.php
drupol: The Filter's callback? If this is it, indeed, I forgot the type hints, I will fix later, after family lunch :)

PR #168, opened 2021-07-31
refactor: Update`Distinct` and `Duplicate` operations.

drupol:
This PR:

* [x] Align the `Duplicate` operation on `Distinct`
* [x] Update the `Distinct` operation
    The distinct operation was not designed in a lazy way.
    The previous version was storing in an array each pair of key-value from the iterator. This is not how we should do that.
    The previous version was based on `Reduce` and was traversing the whole iterator before going to the next operation.
    This is not really optimal and we should update that.
* [x] Has updated documentation and examples
* [x] The `Duplicate` implementation is ready and I'm happy of it

**Because of the nature of these operations and their side effects (the `$stack` variable), I'm unable to test them completely, especially the `continue 2` statement. Infection will most probably detect that without any trouble (_Basically, there is no need to go through the whole `$seen` array once the inner condition is met_).
I haven't found a better way to achieve this, maybe the enlightenment will come another day.**

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -497,7 +497,7 @@ Signature: ``Collection::column($column): Collection;``
 
     $result = Collection::fromIterable($records)
         ->column('first_name'); // ['John', 'Sally', 'Jane', 'Peter']
-    
+
     $result = Collection::fromIterable($records)
         ->column('non_existent_key'); // []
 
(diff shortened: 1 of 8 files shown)


AlexandruGG (approved):
Nice one 👍 .

This would be the kind of refactoring that would really benefit from having [PHPBench](https://phpbench.readthedocs.io/en/latest/index.html) so we could see the difference in runtime and memory used by the operations.

drupol:
> Nice one +1 .
> 
> This would be the kind of refactoring that would really benefit from having [PHPBench](https://phpbench.readthedocs.io/en/latest/index.html) so we could see the difference in runtime and memory used by the operations.

I definitely agree.

That said, I did the tests locally and the difference is significant!

AlexandruGG:
> > Nice one +1 .
> > This would be the kind of refactoring that would really benefit from having [PHPBench](https://phpbench.readthedocs.io/en/latest/index.html) so we could see the difference in runtime and memory used by the operations.
> 
> I definitely agree.
> 
> That said, I did the tests locally and the difference is significant!

I'll find the time at some point to play around with phpbench and see how we introduce it into this project

PR #174, opened 2021-08-04
Documentation: Collection Principles

AlexandruGG:
This PR adds a new section in the documentation to describe a few principles that are at the core of this package.

Also:
* [x] Update docs version in preparation for the release (5.0.0)

Related to https://github.com/loophp/collection/pull/106#issuecomment-860266173.


AlexandruGG:
@drupol hey! There are a few more things which I want to add here, but feel free to have a read of what's there so far and let me know what you think 😁

drupol:
Heya!

I'll make a global feedback at the end, but from what I've seen so far, it's great ! 

Have a great night

docs/pages/principles.rst
```diff
@@ -0,0 +1,89 @@
...
+Most :ref:`operations <Methods (operations)>` return a new collection object
+rather than modifying the original one. This makes your code easier to reason
+about and more testable.
+
+A few notable exceptions to this are methods which return scalar values like ``count``;
+you can read more about them in the :ref:`operations background <Background>` section. 
+
+Laziness
+--------
+
```
drupol, line 19: `Collection` is "lazy" by default, it's one of it's main principle.
Lazy evaluation, or call-by-need, is an evaluation strategy which delays the evaluation of an expression until its value is really needed.
In `Collection`, the operations are executed on the input stream only when iterating over it or when using very specific operation like: `all` or `squash`.

docs/pages/principles.rst
```diff
@@ -0,0 +1,89 @@
...
+with the `invoke`_ PHP magic method. Operations return a ``Closure`` which itself
+returns an ``Iterator``; they can thus be used inside ``ClosureIterator`` to 
+create new generators when needed.
+
+Side-Effects
+~~~~~~~~~~~~
+
+Sometimes you might want your ``Collection`` object to behave eagerly.
+This is a legitimate need when you want to trigger side-effects, like persisting
+entities to a database or sending a batch of emails.
```
drupol, line 67: haha ! People I work with are using `Collection` for doing things like that. As I'm doing the code reviews most of the time, I always ask them to use a regular `foreach` loop instead.
Why?
Because I ultimately think that using `Collection` for doing tasks with side-effects is not the proper way to use this tool.
According to me, if you use `Collection` without using it's return value, then you're using it wrong.
I'm definitely aware that technically it will work, but semantically, it's worse than using a plain old `foreach` (_or alike_) loop.

docs/pages/principles.rst
```diff
@@ -0,0 +1,89 @@
...
+
+This section describes a few elements which are at the core of ``Collection``, 
+and provides guidance on how to get the most out of the package.
+
+Immutability
+------------
+
+Most :ref:`operations <Methods (operations)>` return a new collection object
+rather than modifying the original one. This makes your code easier to reason
+about and more testable.
```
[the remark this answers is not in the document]
AlexandruGG: sounds good to me 👍

docs/pages/principles.rst
AlexandruGG: is this a comment meant to modify what's in the **Behaviour** section, or is this something to include in addition to that?

docs/pages/principles.rst
AlexandruGG: hmm I get what you mean, I guess I'm just not a big fan of loops in general 😅 .

I see `Collection::apply` as a replacement for a separate `foreach` loop (ofc you have to use `squash` as well), because to me doing `apply` + `foreach` seems redundant.

In Rust, iterators have a [for_each](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each) method which basically acts like `apply` but is _not_ lazy, which means it can be a substitute for a loop when you want to do some side effects, especially at the end of a chain of other transformations: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3bcb45d0c7f99d3a91e98e65cf87c371

AlexandruGG: This being said, I'll add a note that normal iteration via a `foreach` loop should be preferred for triggering side-effects

docs/pages/principles.rst
drupol: This was initially for the "Laziness" paragraph, but I noticed that it was kinda the same in the next paragraph. Should we merge them together?

docs/pages/principles.rst
drupol: Ok I see, thanks for the Rust example.

Maybe add a note in this direction:

`Collection` is a helper for making transformations to input data. Despite the fact that you can technically do side-effects in some operations(_due to the fact that you can pass a custom Closure that can alter variables out of its scope_) , it's better to avoid using the library for such endeavor, but rather to get the return value of a transformation.

docs/pages/principles.rst
AlexandruGG: yep I'll see what I can do to merge them

docs/pages/principles.rst
AlexandruGG: @drupol thank you for the suggestion, I've updated the copy now.

I'm still conflicted about the `apply` operation: maybe I'm looking at it the wrong way, but what is it's purpose if not to replace a `foreach` loop? I see it as the equivalent of a loop and similar to Laravel Collection's [each](https://laravel.com/docs/8.x/collections#method-each) method.

drupol: There's no conflict, no worries. `Apply` has been made for that, but I don't advise to use it at all, that's it :)

drupol:
I'm wondering if we should have a word about performance. WDYT ?

I also updated the upcoming changelog and added a sentence about this new documentation page.

AlexandruGG:
> I'm wondering if we should have a word about performance. WDYT ?
> 
> I also updated the upcoming changelog and added a sentence about this new documentation page.

What are you thinking to mention regarding performace?

drupol:
> > I'm wondering if we should have a word about performance. WDYT ?
> > I also updated the upcoming changelog and added a sentence about this new documentation page.
> 
> What are you thinking to mention regarding performace?

That sometimes it might be slower using `Collection` than using simple loops and stuff, but using `Collection` add a semantical value to the code. And despite the fact that it might be slower sometimes, it adds a huge value in terms of readability and maintainability in the long term.

AlexandruGG:
> > > I'm wondering if we should have a word about performance. WDYT ?
> > > I also updated the upcoming changelog and added a sentence about this new documentation page.
> > 
> > 
> > What are you thinking to mention regarding performace?
> 
> That sometimes it might be slower using `Collection` than using simple loops and stuff, but using `Collection` add a semantical value to the code. And despite the fact that it might be slower sometimes, it adds a huge value in terms of readability and maintainability in the long term.

Hmm I'm not sure if this is the best place for think. But this gave me an idea for a separate section I can add in the future, where this would be a great mention. This section would compare how to do the same thing:
1. the "imperative" way
2. using Collection or Operations

docs/pages/code/operations/same.php
```diff
@@ -48,14 +48,14 @@
...
 
 $comparator = static fn ($left, $leftKey) => static fn ($right, $rightKey): bool => $left === $right
     && mb_strtolower($leftKey) === mb_strtolower($rightKey);
-$this::fromIterable(['foo' => 'f'])
+Collection::fromIterable(['foo' => 'f'])
     ->same(Collection::fromIterable(['FOO' => 'f']), $comparator); // true
 
 $comparator = static fn (stdClass $left) => static fn (stdClass $right): bool => $left->id === $right->id;
-$this::fromIterable([$a])
+Collection::fromIterable([$a])
```
AlexandruGG, line 60: bad copy-paste here 😅

AlexandruGG:
@drupol I added the remaining sections I wanted now. Have a read through and let me know if it makes sense and any feedback you might have 😁

docs/pages/principles.rst
```diff
@@ -0,0 +1,175 @@
...
+For example, the :ref:`contains <Contains>` operation allows us to easily check whether one or more
+values are contained in the collection:
+
+.. code-block:: php
+
+    $result = Collection::fromIterable(range('a', 'c'))
+        ->contains('d'); // false
+
+    $result = Collection::fromIterable(range('a', 'c'))
+        ->contains('a', 'z'); // true
```
drupol, line 130: How about adding an example with a AND and OR using filter?

AlexandruGG: in addition to this or instead of this example?

drupol: In addition 👍

AlexandruGG: done!

AlexandruGG:
@drupol anything else on this? Otherwise it's ready from my side 🚀

drupol (approved):
All good !

PR #175, opened 2021-08-05
Refactor/update `Span` and `Partition` in point-free

drupol:
This PR:

* [x] Rewrite `Partition` operation in point free style and without local variables.
* [x] Rewrite `Span` operation in point free style and without local variables.
* [x] Fix typing information of `TakeWhile` and `DropWhile`

### DIFF
--- a/src/Operation/Partition.php
+++ b/src/Operation/Partition.php
@@ -39,7 +39,7 @@ final class Partition extends AbstractOperation
                 /**
                  * @param callable(T, TKey, Iterator<TKey, T>): bool ...$callbacks
                  */
-                static fn (callable ...$callbacks) =>
+                static fn (callable ...$callbacks): Closure =>
                 /**
                  * @param Iterator<TKey, T> $iterator
                  *
--- a/src/Operation/Partition.php
+++ b/src/Operation/Partition.php
@@ -34,10 +34,14 @@ final class Partition extends AbstractOperation
             (
                 /**
                  * @param array{0:(Closure(callable(T, TKey, Iterator<TKey, T>): bool ...): (Closure(Iterator<TKey, T>): Generator<TKey, T>)), 1:(Closure(callable(T, TKey, Iterator<TKey, T>): bool ...): (Closure(Iterator<TKey, T>): Generator<TKey, T>))} $operations
+                 *
+                 * @return Closure(callable(T, TKey, Iterator<TKey, T>): bool ...): Closure(Iterator<TKey, T>): Generator<int, array{0: Closure(Iterator<TKey, T>): Generator<TKey, T>, 1: array{0: Iterator<TKey, T>}}>
                  */
                 static fn (array $operations): Closure =>
                 /**
                  * @param callable(T, TKey, Iterator<TKey, T>): bool ...$callbacks
+                 *
+                 * @return Closure(Iterator<TKey, T>): Generator<int, array{0: Closure(Iterator<TKey, T>): Generator<TKey, T>, 1: array{0: Iterator<TKey, T>}}>
                  */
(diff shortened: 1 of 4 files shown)


src/Operation/Span.php
```diff
@@ -31,25 +31,29 @@ final class Span extends AbstractOperation
...
-            static fn (callable ...$callbacks): Closure =>
+            (
+                /**
+                 * @param array{0:(Closure(callable(T, TKey, Iterator<TKey, T>): bool ...): (Closure(Iterator<TKey, T>): Generator<TKey, T>)), 1:(Closure(callable(T, TKey, Iterator<TKey, T>): bool ...): (Closure(Iterator<TKey, T>): Generator<TKey, T>))} $operations
+                 */
+                static fn (array $operations): Closure =>
+                /**
+                 * @param callable(T, TKey, Iterator<TKey, T>): bool ...$callbacks
+                 */
+                static fn (callable ...$callbacks) =>
```
AlexandruGG, line 42: what happened to the `Closure` return type here?

src/Operation/Partition.php
```diff
@@ -31,25 +31,29 @@ final class Partition extends AbstractOperation
...
-            static fn (callable ...$callbacks): Closure =>
+            (
+                /**
+                 * @param array{0:(Closure(callable(T, TKey, Iterator<TKey, T>): bool ...): (Closure(Iterator<TKey, T>): Generator<TKey, T>)), 1:(Closure(callable(T, TKey, Iterator<TKey, T>): bool ...): (Closure(Iterator<TKey, T>): Generator<TKey, T>))} $operations
+                 */
+                static fn (array $operations): Closure =>
+                /**
+                 * @param callable(T, TKey, Iterator<TKey, T>): bool ...$callbacks
+                 */
+                static fn (callable ...$callbacks) =>
```
AlexandruGG, line 42: same here, I don't think we should remove the return type

src/Operation/Span.php
drupol: Oops! it's fixed now.

src/Operation/Partition.php
drupol: Fixed as well

src/Operation/Partition.php
```diff
@@ -31,25 +31,29 @@ final class Partition extends AbstractOperation
     public function __invoke(): Closure
     {
         return
-            /**
-             * @param callable(T, TKey, Iterator<TKey, T>): bool ...$callbacks
-             *
-             * @return Closure(Iterator<TKey, T>): Generator<int, array{0: Closure(Iterator<TKey, T>): Generator<TKey, T>, 1: array{0: Iterator<TKey, T>}}>
```
AlexandruGG, line 37: still missing this Closure return type which was removed

src/Operation/Span.php
```diff
@@ -31,25 +31,29 @@ final class Span extends AbstractOperation
     public function __invoke(): Closure
     {
         return
-            /**
-             * @param callable(T, TKey, Iterator<TKey, T>): bool ...$callbacks
-             *
-             * @return Closure(Iterator<TKey, T>): Generator<int, array{0: Closure(Iterator<TKey, T>): Generator<TKey, T>, 1: array{0: Iterator<TKey, T>}}>
```
AlexandruGG, line 37: same here

src/Operation/Partition.php
drupol: I added them back despite the fact that they were not really required.

src/Operation/Span.php
drupol: Done.

AlexandruGG (approved):

PR #177, opened 2021-08-15
refactor: Update cross `product` operation in point free style.

drupol:
This PR:

* [x] Update `Product` in point free style.

### DIFF
--- a/spec/loophp/collection/CollectionSpec.php
+++ b/spec/loophp/collection/CollectionSpec.php
@@ -360,6 +360,14 @@ class CollectionSpec extends ObjectBehavior
         $this::fromResource($stream)
             ->shouldThrow(InvalidArgumentException::class)
             ->during('all');
+
+        $string = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit.';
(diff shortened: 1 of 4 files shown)


drupol:
It seems that there is an issue with the latest PSalm version (4.9.3)... investigating.

drupol:
I updated `drupol/php-conventions` so it doesn't install version 4.9.3 of Psalm.

AlexandruGG (approved):
I cannot see the Psalm issue anymore but looks like you sorted it by restricting the version so should be fine for now.

On a separate note, I think this new form is extremely difficult to follow and understand 😅.
What do you think of using some variables just to make it a bit easier to follow? It shouldn't affect performance too much.

For example, storing [this closure](https://github.com/loophp/collection/blob/91535c35d6c341c8bf8b313a410a5fa00636622c/src/Operation/Product.php#L43-L62) in a variable. Same for the others, and then combining in the Pipe. Alternatively, we could add comments to show what each one represents?

AlexandruGG:
@drupol auto-merge got me again 😄

PR #178, opened 2021-08-15
refactor: Update `Transpose` operation in point free style.

drupol:
This PR:

* [x] Update `Transpose` in point free style.

### DIFF
--- a/src/Operation/Transpose.php
+++ b/src/Operation/Transpose.php
@@ -26,6 +26,8 @@ final class Transpose extends AbstractOperation
     /**
      * @pure
      *
+     * @psalm-suppress ImpureMethodCall
+     *
      * @return Closure(Iterator<TKey, T>): Generator<TKey, list<T>>
      */
     public function __invoke(): Closure
--- a/src/Operation/Transpose.php
+++ b/src/Operation/Transpose.php
@@ -30,42 +30,39 @@ final class Transpose extends AbstractOperation
      */
     public function __invoke(): Closure
     {
-        return
+        $callbackForKeys =
             /**
-             * @param Iterator<TKey, T> $iterator
+             * @param array $carry
+             * @param non-empty-array<int, TKey> $key
              *
-             * @return Generator<TKey, list<T>>
+             * @return TKey
              */
-            static function (Iterator $iterator): Generator {
-                $mit = new MultipleIterator(MultipleIterator::MIT_NEED_ANY);
+            static fn (array $carry, array $key) => reset($key);
 
-                foreach ($iterator as $iteratorIterator) {
-                    $mit->attachIterator(new IterableIterator($iteratorIterator));
+        $callbackForValues =
+            /**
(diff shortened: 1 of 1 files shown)


src/Operation/Transpose.php
```diff
@@ -30,42 +30,39 @@ final class Transpose extends AbstractOperation
...
             /**
-             * @param Iterator<TKey, T> $iterator
+             * @param array $carry
+             * @param non-empty-array<int, TKey> $key
              *
-             * @return Generator<TKey, list<T>>
+             * @return TKey
              */
-            static function (Iterator $iterator): Generator {
-                $mit = new MultipleIterator(MultipleIterator::MIT_NEED_ANY);
```
AlexandruGG, line 40: hmm removing this closure makes the operation impure now because of the `MultipleIterator` instantiation. Before this instantiation happened in this returned closure, which meant that the operation itself was still pure, even though it returned an impure closure.

Is there any way we can keep this behaviour in the refactor?

drupol: Indeed. I asked the question on Psalm's slack channel and they told me to use a stub. I tried, but I'm going nowhere.
I wish I could find a way to avoid adding false-positives... but.

However, in this particular case I think we can safely ignore the issue.

1. First, because the Transpose operation will never return a `MultipleIterator`.
2. Second, because the Transpose operation is using a `MultipleIterator` object to determine the result. This object is an intermediary variable(tool) that is used to find the resulting solution.

I will set this PR as draft in the meantime, I'll try to come up with a better solution.

drupol: I quickly drafted a simple example here: https://3v4l.org/LNuDT

Despite the fact that I use a `MultipleIterator`, this is completely pure.

AlexandruGG: @drupol what I meant here is this: can we keep the same behaviour where we use the `MultipleIterator` in a Closure? I know it's not really an issue and it's safe to ignore, I'm just wondering if it can be done.

Something like:

```PHP
{
    return 
        static function () {
            $pipe = ... // this might have to be a bit different

            return $pipe; // might need to call it here e.g. $pipe();
        }
}

Comment (drupol) [reply src/Operation/Transpose.php:40]: Going to do a bit more research to fulfill your request and I'll let you know asap.

File: src/Operation/Transpose.php [hunk 2]
```diff
@@ -26,46 +26,45 @@ final class Transpose extends AbstractOperation
     /**
      * @pure
      *
+     * @psalm-suppress ImpureMethodCall
```
AlexandruGG, line 29: if we do end up keeping this we should always have clarifying comments for them, e.g.

```
@psalm-suppress ImpureMethodCall - using MultipleIterator as an internal tool which is not returned

drupol: I just see this message! Good point, going to make the change.

PR #179, opened 2021-08-15
refactor: Update `Window` operation in point free style.

drupol:
This PR:

* [x] Update `Window` in point free style.
* [x] Has updated tests
* [x] Is a minor BC break - The behavior has changed when the ```$size``` is zero, which is more in sync with what the Window operation does and avoid mixing types (```T|list<T>``` => ```list<T>```)
  Before: ```Collection::fromIterable(['a', 'b', 'c'])->window(0); // ['a', 'b', 'c']```
  After:  ```Collection::fromIterable(['a', 'b', 'c'])->window(0); // [['a'], ['b'], ['c']]```

### DIFF
--- a/spec/loophp/collection/CollectionSpec.php
+++ b/spec/loophp/collection/CollectionSpec.php
@@ -4012,15 +4012,15 @@ class CollectionSpec extends ObjectBehavior
 
     public function it_can_window(): void
     {
-        $this::fromIterable(range('a', 'c'))
+        $this::fromIterable(['a' => 'A', 'b' => 'B', 'c' => 'C'])
             ->window(0)
             ->shouldIterateAs([
-                ['a'],
-                ['b'],
-                ['c'],
+                'a' => ['A'],
+                'b' => ['B'],
+                'c' => ['C'],
             ]);
 
-        $this::fromIterable(range('a', 'z'))
+        $this::fromIterable(range('a', 'e'))
             ->window(2)
             ->shouldIterateAs([
                 0 => [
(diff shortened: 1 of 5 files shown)


spec/loophp/collection/CollectionSpec.php
```diff
@@ -3999,9 +3999,13 @@ public function it_can_when(): void
...
     {
-        $this::fromIterable(range('a', 'z'))
+        $this::fromIterable(range('a', 'c'))
             ->window(0)
-            ->shouldIterateAs(range('a', 'z'));
+            ->shouldIterateAs([
+                ['a'],
+                ['b'],
+                ['c'],
+            ]);
```
AlexandruGG, line 4008: is this change a BC-break? Before `window(0)` would return the items essentially unchanged, but now it puts every element in an array

drupol: Right, this is indeed a minor BC Break. We can even pass negative numbers and get a different behavior that might be useful. I completely forgot to document that. Thanks.

spec/loophp/collection/CollectionSpec.php
```diff
@@ -4041,111 +4045,13 @@ public function it_can_window(): void
...
-                25 => [
-                    0 => 'x',
-                    1 => 'y',
-                    2 => 'z',
-                ],
+            ]);
+
+        // Unsupported - but tested.
+        $this::fromIterable(range('a', 'e'))
+            ->window(-2)
```
AlexandruGG, line 4052: what about the `window(-1)` case? That one is mentioned now in the docs so sounds like it's supported -> we should have a test for it

AlexandruGG:
@drupol I think you merged the wrong one? 😅

drupol:
Oh. Sorry :( Indeed I messed up... too many tabs open here.

drupol:
I'll add a new test in the master branch directly in a few minutes.

PR #181, opened 2021-08-15
refactor: Update `Zip` operation in point free style.

drupol:
This PR:

* [x] Update `Zip` in point free style.

### DIFF
--- a/src/Contract/Operation/Zipable.php
+++ b/src/Contract/Operation/Zipable.php
@@ -24,7 +24,7 @@ interface Zipable
      *
      * @param iterable<mixed, mixed> ...$iterables
      *
-     * @return Collection<list<mixed|TKey>, list<mixed|T>>
+     * @return Collection<list<mixed>, list<mixed>>
      */
     public function zip(iterable ...$iterables): Collection;
 }
--- a/src/Contract/Operation/Zipable.php
+++ b/src/Contract/Operation/Zipable.php
@@ -22,9 +22,9 @@ interface Zipable
      *
      * @see https://loophp-collection.readthedocs.io/en/stable/pages/api.html#zip
      *
-     * @param iterable<mixed> ...$iterables
+     * @param iterable<mixed, mixed> ...$iterables
      *
-     * @return Collection<TKey,T>
+     * @return Collection<list<mixed|TKey>, list<mixed|T>>
      */
     public function zip(iterable ...$iterables): Collection;
 }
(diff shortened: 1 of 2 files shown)


AlexandruGG (commented):
@drupol this one feels like another case where things are increasing in complexity and more dependencies added (Pipe, Reduce, Flatten), especially if we compare to the previous form which was relatively simple.

What's the benefit of doing this? And is there any way to make things a bit more readable?

drupol:
> The point-free style is considered by some author as the ultimate functional programming style.
> 
> To put things simply, a function of type ```t1 -> t2``` describes a transformation from one element of type ```t1``` into another element of type ```t2```. The idea is that "pointful" functions (_written using variables_) emphasize elements (when you write ```function (Iterator $iterator) => ...```, you're describing what's happening to the element ```$iterator```), while "point-free" functions (_expressed without using variables_) emphasize the transformation itself, as a **composition** of simpler transforms. 
> 
> Advocates of the point-free style argue that transformations should indeed be the central concept, and that the pointful notation, while easy to use, distracts us from this noble ideal.
> 

This library is about **transformations**, and I think we should emphasize that in the code.

I also consider this as a challenge in order to be succinct and to express pipelined computations as a composition of functions rather than thinking of threading arguments through each other.

Actually I found this very readable, it's just a matter of habits, that said, I will make a slight modification tonight that will make it a bit more easier to read :)

src/Contract/Operation/Zipable.php
```diff
@@ -22,9 +22,9 @@ interface Zipable
      *
      * @see https://loophp-collection.readthedocs.io/en/stable/pages/api.html#zip
      *
-     * @param iterable<mixed> ...$iterables
+     * @param iterable<mixed, mixed> ...$iterables
      *
-     * @return Collection<TKey,T>
+     * @return Collection<list<mixed|TKey>, list<mixed|T>>
```
AlexandruGG, line 27:
```suggestion
     * @return Collection<list<mixed>, list<mixed>>
```

`mixed` already covers TKey and T, so this could be simplified as such.

AlexandruGG (commented):
@drupol Thanks for the explanation and the little refactor! 

I completely understand your view of the transformations and the beauty of composing functions to achieve them. I think the new form is definitely clearer in what it's doing 😁 

Feel free to merge this, I'm not hitting "approve" so you get the chance to check the other comment 😄

PR #186, opened 2021-08-23
refactor: Update `Init` operation in point free style.

drupol:
This PR:

* [x] Update `Init` in point free style.
* [x] Includes a minor out of scope change in `TakeWhile` operation.

### DIFF
--- a/src/Operation/Init.php
+++ b/src/Operation/Init.php
@@ -37,7 +37,7 @@ final class Init extends AbstractOperation
              * @param TKey $key
              * @param CachingIterator<TKey, T> $iterator
              */
-            static fn ($value, $key, CachingIterator $iterator): bool => $iterator->getInnerIterator()->valid();
+            static fn ($value, $key, CachingIterator $iterator): bool => $iterator->hasNext();
 
         $buildCachingIterator =
             /**
--- a/src/Operation/Init.php
+++ b/src/Operation/Init.php
@@ -19,6 +19,8 @@ use Iterator;
  *
  * @template TKey
  * @template T
+ *
+ * phpcs:disable Generic.Files.LineLength.TooLong
  */
 final class Init extends AbstractOperation
 {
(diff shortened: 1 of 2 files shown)


src/Operation/Init.php
```diff
@@ -29,22 +31,29 @@ final class Init extends AbstractOperation
...
     public function __invoke(): Closure
     {
-        return
+        $callback =
+            /**
+             * @param T $value
+             * @param TKey $key
+             * @param CachingIterator<TKey, T> $iterator
+             */
+            static fn ($value, $key, CachingIterator $iterator): bool => $iterator->getInnerIterator()->valid();
```
AlexandruGG, line 40: for some reason Psalm thinks that `getInnerIterator` returns `Traversable` here instead of `Iterator`, but I cannot replicate it in the playground: https://psalm.dev/r/037fdbb040. The trace shows `Iterator` there.

But in this project:
```
INFO: Trace - src/Operation/Init.php:42:17 - $inner: Traversable (see https://psalm.dev/224)
                /** @psalm-trace $inner */
                $inner = $iterator->getInnerIterator();
```

Could it be the Psalm version? (we must be using an older one)

drupol: Haha ! I almost made the same Psalm example to try it out on my own :)

I don't know where the issue comes from yet, I'm investigating!

drupol: It only happens with PHP 8. Weird.

drupol: I think I should submit this to PSalm.

image

drupol: Weird. Using `CachingIterator::hasNext()` fixed the issue.

AlexandruGG: I think we should open an issue on Psalm with that, sounds like a problem in PHP 8

AlexandruGG (approved):

src/Operation/Init.php
drupol: I did here: https://github.com/vimeo/psalm/pull/6359 we'll see how it goes.

PR #192, opened 2021-08-27
Static analysis: Update `Map` operation.

drupol:
This PR:

* [ ] Fix `Map` operation

Psalm snippets:

* Original one (current state): https://psalm.dev/r/9d647acb4e
* Alex's idea: https://psalm.dev/r/e840743851
  * Pro: Typing is done properly
  * Cons: It requires changing the whole API
* Pol's idea: https://psalm.dev/r/bf2e69c00a
  * Pro: It does not require changing the whole API
  * Cons: So sad, it seems [to not be supported by PSalm](https://github.com/vimeo/psalm/issues/5472), but it is by PHPStan: https://phpstan.org/r/1f5caae6-4b6d-4e98-bcd7-cd7ba9d546e0

In the end, I'm here (https://psalm.dev/r/33088712ff), there's only one issue remaining.

I asked @weirdan and he proposed:
   * https://psalm.dev/r/37900dd56d
   * https://psalm.dev/r/3930c24012

I'm still experimenting on the best way to fix this issue.

### DIFF
--- a/src/Contract/Collection.php
+++ b/src/Contract/Collection.php
@@ -134,6 +134,7 @@ use loophp\collection\Contract\Operation\Zipable;
  *
  * @template TKey
  * @template T
+ * @template V
  *
  * @template-extends Allable<TKey, T>
  * @template-extends Appendable<TKey, T>
@@ -192,7 +193,7 @@ use loophp\collection\Contract\Operation\Zipable;
  * @template-extends Lastable<TKey, T>
  * @template-extends Limitable<TKey, T>
  * @template-extends Linesable<TKey, T>
- * @template-extends Mapable<TKey, T>
+ * @template-extends Mapable<TKey, T, V>
  * @template-extends MapNable<TKey, T>
  * @template-extends Matchable<TKey, T>
  * @template-extends Matchingable<TKey, T>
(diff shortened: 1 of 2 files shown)


src/Contract/Collection.php
```diff
@@ -192,7 +193,7 @@
  * @template-extends Lastable<TKey, T>
  * @template-extends Limitable<TKey, T>
  * @template-extends Linesable<TKey, T>
- * @template-extends Mapable<TKey, T>
+ * @template-extends Mapable<TKey, T, V>
```
AlexandruGG, line 196: seeing the errors on this I'm thinking it's not the right way.

Looking at the [Doctrine ArrayCollection](https://github.com/doctrine/collections/blob/1.6.x/lib/Doctrine/Common/Collections/ArrayCollection.php#L347) the template variable is defined on the method.

I played with the Psalm example a bit more and came up with another version which works while keeping the template param at the same level:
https://psalm.dev/r/e840743851
https://phpstan.org/r/5166ee20-1783-4d1d-93f6-91e4ac3570b4

Note the callback is passed from the invoke method. I think the issue with the initial one might be that the template is defined at a different level from where the callback is passed, so it "can't know" the return type of the callback. This is just my assumption, I can't claim to understand this fully 😄

drupol: Going to analyze this later in the day for sure!

drupol: Voila, I updated the top message with the relevant PSalm snippets.

I'm still searching for the best way to implement this.

PR #194, opened 2021-09-09
refactor: Fix PSalm issues.

drupol:
This PR:

* [x] Fix PSalm issues

I'm definitely unsure about this thing, @AlexandruGG WDYT?

### DIFF
--- a/tests/static-analysis/first.php
+++ b/tests/static-analysis/first.php
@@ -53,9 +53,9 @@ first_checkStringElement(Collection::fromIterable(['foo' => 'bar', 'baz' => 'bar
 first_checkStringElement(Collection::fromIterable(['foo' => 'bar', 'baz' => 'bar'])->first()->all()['baz']);
 
 // VALID failures - `current` returns T|null
-/** @psalm-suppress PossiblyNullArgument @phpstan-ignore-next-line */
+/** @psalm-suppress NullArgument @phpstan-ignore-next-line */
 first_checkIntElement(Collection::fromIterable([1, 2, 3])->first()->current());
-/** @psalm-suppress PossiblyNullArgument @phpstan-ignore-next-line */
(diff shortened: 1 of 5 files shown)


tests/static-analysis/first.php
```diff
@@ -53,9 +53,9 @@ function first_checkNullableString(?string $value): void
 first_checkStringElement(Collection::fromIterable(['foo' => 'bar', 'baz' => 'bar'])->first()->all()['baz']);
 
 // VALID failures - `current` returns T|null
-/** @psalm-suppress PossiblyNullArgument @phpstan-ignore-next-line */
+/** @psalm-suppress NullArgument @phpstan-ignore-next-line */
 first_checkIntElement(Collection::fromIterable([1, 2, 3])->first()->current());
-/** @psalm-suppress PossiblyNullArgument @phpstan-ignore-next-line */
+/** @psalm-suppress NullArgument @phpstan-ignore-next-line */
```
AlexandruGG, line 58: this might be problematic I think if Psalm always thinks that these are `null`. I tried debugging a bit:

```
INFO: Trace - tests/static-analysis/last.php:59:1 - $var: loophp\collection\Contract\Collection<0|1|2, 1|2|3>&loophp\collection\Contract\Collection (see https://psalm.dev/224)
/** @psalm-trace $var */
$var = Collection::fromIterable([1, 2, 3])->first();
```

Psalm now sees `first()` as creating an intersection type for some reason and I think that's why when applying `current()` after it resolves the type to `null`.

I tried to replicate somewhat in the playground but couldn't :( https://psalm.dev/r/609938fcc6

AlexandruGG (approved):
The code changes themselves are fine for now so I think there's no harm in getting this merged to fix the code analysis. We might need to figure out the root of the cause for this later, though it might just be a Psalm bug.

PR #196, opened 2021-09-13
Update various typing information

drupol:
This PR:

* [x] The `Associate` class is no more variadic - (Minor BC Break - only when using the `Associate` class on its own)
* [x] Fix `Zip` and `Associate` typing information
* [x] Update `Unpack` operation and its interface. 
* [x] `Associate` documentation/examples has been updated - added more examples.

### DIFF
--- a/phpstan-unsupported-baseline.neon
+++ b/phpstan-unsupported-baseline.neon
@@ -1,11 +1,6 @@
 parameters:
 	ignoreErrors:
 		-
-			message: "#^Parameter \\#1 \\$callable of class loophp\\\\collection\\\\Collection constructor expects callable\\(\\.\\.\\.mixed\\)\\: iterable\\<int, string\\>, Closure\\(string, string\\)\\: loophp\\\\collection\\\\Iterator\\\\StringIterator\\<mixed, string\\> given\\.$#"
-			count: 1
-			path: src/Collection.php
-
-		-
 			message: "#^Generic type loophp\\\\collection\\\\Contract\\\\Operation\\\\Unpackable\\<mixed, array\\<int, mixed\\>\\> in PHPDoc tag @extends does not specify all template types of interface loophp\\\\collection\\\\Contract\\\\Operation\\\\Unpackable\\: TKey, T, NewTKey, NewT$#"
 			count: 1
 			path: src/Contract/Collection.php
@@ -21,17 +16,6 @@ parameters:
 			path: src/Contract/Operation/Unpackable.php
 
 		-
-			message: "#^Parameter \\#1 \\$callable of class loophp\\\\collection\\\\Iterator\\\\ClosureIterator constructor expects callable\\(\\.\\.\\.mixed\\)\\: iterable\\<int, string\\>, Closure\\(mixed, bool\\)\\: Generator\\<int, string, mixed, void\\> given\\.$#"
-			count: 1
-			path: src/Iterator/ResourceIterator.php
-
-		-
-			message: "#^Parameter \\#1 \\$callable of class loophp\\\\collection\\\\Iterator\\\\ClosureIterator constructor expects callable\\(\\.\\.\\.mixed\\)\\: iterable, Closure\\(Iterator, callable\\(\\)\\: mixed\\)\\: Generator\\<mixed, mixed, mixed, void\\> given\\.$#"
-			count: 1
-			path: src/Iterator/TypedIterator.php
-
-		-
 			message: "#^PHPDoc tag @template T for class loophp\\\\collection\\\\Operation\\\\Unpack with bound type array\\<int, NewT\\|NewTKey\\> is not supported\\.$#"
 			count: 1
 			path: src/Operation/Unpack.php
-
--- a/phpstan-unsupported-baseline.neon
+++ b/phpstan-unsupported-baseline.neon
@@ -6,7 +6,7 @@ parameters:
 			path: src/Collection.php
 
 		-
-			message: "#^Generic type loophp\\\\collection\\\\Contract\\\\Operation\\\\Unpackable\\<mixed, mixed\\> in PHPDoc tag @extends does not specify all template types of interface loophp\\\\collection\\\\Contract\\\\Operation\\\\Unpackable\\: TKey, T, NewTKey, NewT$#"
+			message: "#^Generic type loophp\\\\collection\\\\Contract\\\\Operation\\\\Unpackable\\<mixed, array\\<int, mixed\\>\\> in PHPDoc tag @extends does not specify all template types of interface loophp\\\\collection\\\\Contract\\\\Operation\\\\Unpackable\\: TKey, T, NewTKey, NewT$#"
 			count: 1
 			path: src/Contract/Collection.php
 
@@ -21,6 +21,17 @@ parameters:
 			path: src/Contract/Operation/Unpackable.php
 
 		-
+			message: "#^Parameter \\#1 \\$callable of class loophp\\\\collection\\\\Iterator\\\\ClosureIterator constructor expects callable\\(\\.\\.\\.mixed\\)\\: iterable\\<int, string\\>, Closure\\(mixed, bool\\)\\: Generator\\<int, string, mixed, void\\> given\\.$#"
+			count: 1
+			path: src/Iterator/ResourceIterator.php
+
+		-
+			message: "#^Parameter \\#1 \\$callable of class loophp\\\\collection\\\\Iterator\\\\ClosureIterator constructor expects callable\\(\\.\\.\\.mixed\\)\\: iterable, Closure\\(Iterator, callable\\(\\)\\: mixed\\)\\: Generator\\<mixed, mixed, mixed, void\\> given\\.$#"
+			count: 1
+			path: src/Iterator/TypedIterator.php
+
+		-
 			message: "#^PHPDoc tag @template T for class loophp\\\\collection\\\\Operation\\\\Unpack with bound type array\\<int, NewT\\|NewTKey\\> is not supported\\.$#"
 			count: 1
 			path: src/Operation/Unpack.php
+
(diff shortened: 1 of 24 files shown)


drupol:
Hi @AlexandruGG,

I think this PR could be a nice addition, however, I have 4 issues in PHPStan and 2 in PSalm.

Do you think we could ignore them or you have an idea on how to fix them?

src/Contract/Operation/Zipable.php
```diff
@@ -22,9 +22,12 @@ interface Zipable
      *
      * @see https://loophp-collection.readthedocs.io/en/stable/pages/api.html#zip
      *
-     * @param iterable<mixed, mixed> ...$iterables
+     * @template U
+     * @template UKey
      *
-     * @return Collection<list<mixed>, list<mixed>>
+     * @param iterable<UKey, U> ...$iterables
```
drupol, line 28: I know that it's not possible to type this thing as it is a variadic parameter.

I don't know if it's a good idea, but I had the idea to do the typing of only one element, knowing that it would not be completely valid if there are more than one parameters given.

WDYT?

drupol:
I fixed the issue with PSalm and PHPStan, this is not ready for review.

src/Contract/Operation/Unpackable.php
```diff
@@ -13,10 +13,7 @@
 
 /**
  * @template TKey
- * @template T of array{0: NewTKey, 1: NewT}
- *
- * @template NewTKey
- * @template NewT
+ * @template T
```
AlexandruGG, line 16: I don't get this change. Was the existing annotation not working? Or is it not correct?

There's no static analysis check for `unpack` so this would be the time to add one if we're changing these

src/Contract/Operation/Zipable.php
AlexandruGG: I get what you're looking to do here. The only issue is that when a user will try to use it with multiple iterables of different types they will get an error and so think that it's not supposed to be used like that.

I'm not sure how people normally use `zip`, maybe it wouldn't be an issue if most often it's used with a single type.

src/Operation/Normalize.php
```diff
@@ -24,15 +24,15 @@ final class Normalize extends AbstractOperation
...
+     * @return Closure(Iterator<TKey, T>): Generator<int, T, mixed, void>
      */
     public function __invoke(): Closure
     {
         return
             /**
              * @param Iterator<TKey, T> $iterator
              *
-             * @return Generator<int, T>
+             * @return Generator<int, T, mixed, void>
```
AlexandruGG, line 35: does it add to the type information to add the other Generator generics, or were they already inferred as those types?

tests/static-analysis/zip.php
```diff
@@ -0,0 +1,43 @@
...
+     */
+    static function (): Generator {
+        yield true => true;
+
+        yield false => false;
+
+        yield true => true;
+    };
+
+zip_checkBoolString(Collection::fromIterable($generator())->zip(range('a', 'c')));
```
AlexandruGG, line 43: we should have a few more checks:
- zipping with only 1 iterable which is of the same type as the initial collection
- zipping with multiple iterables which are of the same type as the initial collection
- zipping with multiple iterables which are of different types as the initial collections -> here there should be errors which we need to ignore

phpstan-unsupported-baseline.neon
```diff
@@ -1,26 +1,17 @@
 parameters:
 	ignoreErrors:
 		-
-			message: "#^Parameter \\#1 \\$callable of class loophp\\\\collection\\\\Collection constructor expects callable\\(\\.\\.\\.mixed\\)\\: iterable\\<int, string\\>, Closure\\(string, string\\)\\: loophp\\\\collection\\\\Iterator\\\\StringIterator\\<mixed, string\\> given\\.$#"
+			message: "#^Method loophp\\\\collection\\\\Collection\\:\\:squash\\(\\) should return loophp\\\\collection\\\\Contract\\\\Collection\\<TKey, T\\> but returns loophp\\\\collection\\\\Contract\\\\Collection\\<int, array\\(TKey, T\\)\\>\\.$#"
```
AlexandruGG, line 4: this error is new and my bet is it's due to the type updates in `unpack`

src/Operation/Normalize.php
drupol: Hi Alex,

I don't understand your question, can you elaborate?

To give you a bit of context, this was spotted by PHPStan when using (see [this tweet](https://twitter.com/OndrejMirtes/status/1437291613336784902)):

```yaml
parameters:
    checkExplicitMixed: true
```

This is why I added some of these things here and there.

tests/static-analysis/zip.php
drupol: Yep, I will update.

src/Contract/Operation/Unpackable.php
drupol: Despite the fact that this was technically valid, I changed that because:

1. All operations interfaces are using 2 templates (`TKey` and `T`), `Unpackable` was using 4 (`TKey`, `T`, `NewTKey`, `NewT`)
2. Incoming templates (`TKey` and `T`) should be opaque, at least that's how it works on every other operations. We should not describe the structure of these or else it will never end. If we start to do that, where should we put the limit or depth of details?
3. By removing that, we can remove a bunch of "unsupported" lines from `phpstan-unsupported-baseline.neon`

Regarding the SA for `Unpack`, I will add it today for sure!

phpstan-unsupported-baseline.neon
drupol: Yes, this is a big mystery for me. I don't know where this is coming from yet. I added it in that file, but I'm planning to remove it. I might need some help on this thing if you have time.

src/Contract/Operation/Zipable.php
drupol: I leave this question open as I don't have any idea yet on what is best to do...

src/Contract/Operation/Unpackable.php
drupol: Ok, I reverted the changes, I cannot get through it properly in PHPStan.

phpstan-unsupported-baseline.neon
drupol: This has been fixed now that I reverted the change in `Unpack`.

tests/static-analysis/zip.php
drupol: Job done.

phpstan-unsupported-baseline.neon
```diff
@@ -20,7 +20,18 @@ parameters:
 			count: 1
 			path: src/Contract/Operation/Unpackable.php
 
+		-
+			message: "#^Parameter \\#1 \\$callable of class loophp\\\\collection\\\\Iterator\\\\ClosureIterator constructor expects callable\\(\\.\\.\\.mixed\\)\\: iterable\\<int, string\\>, Closure\\(mixed, bool\\)\\: Generator\\<int, string, mixed, void\\> given\\.$#"
```
drupol, line 24: This is the only part (The 2 new entries in this file) that I cannot fix. If you have any idea, just let me know.

src/Operation/Normalize.php
AlexandruGG: ok, sounds good! Should we add that in the PHPStan config in a separate PR?

drupol: No, not yet, I'll wait the next release of Phpstan before making those changes. I even think that it will be enabled by default in the new level 9 :)

src/Iterator/TypedIterator.php
```diff
@@ -28,7 +28,7 @@ final class TypedIterator extends ProxyIterator
 {
     /**
      * @param Iterator<TKey, T> $iterator
-     * @param null|callable(mixed): string $getType
+     * @param callable(T): string $getType
```
AlexandruGG, line 31:
```suggestion
     * @param null|callable(T): string $getType
```
the callable is nullable here

phpstan-unsupported-baseline.neon
AlexandruGG: it's a bit difficult to tell at a glance, but leave this will me I will try to see if I can figure out the problem and fix it

drupol: Ok, thanks! :)

src/Collection.php
```diff
@@ -508,10 +507,7 @@ static function (Generator $generator): Generator {
      */
     public static function fromIterable(iterable $iterable): self
     {
-        return new self(
-            static fn (iterable $iterable): Iterator => new IterableIterator($iterable),
-            [$iterable]
-        );
+        return new self(static fn (): Iterator => new IterableIterator($iterable));
```
AlexandruGG, line 510: does it matter if we pass the parameter to the closure or we just let it capture it from the environment? At least with the testing we have it doesn't seem to make any difference, but it fixes a PHPStan error

src/Collection.php
```diff
@@ -539,10 +529,7 @@ public static function fromResource($resource): self
      */
     public static function fromString(string $string, string $delimiter = ''): self
     {
-        return new self(
-            static fn (string $string, string $delimiter): Iterator => new StringIterator($string, $delimiter),
-            [$string, $delimiter]
-        );
+        return new self(static fn (): Iterator => new StringIterator($string, $delimiter));
```
AlexandruGG, line 532: does it matter if we pass the parameter to the closure or we just let it capture it from the environment? At least with the testing we have it doesn't seem to make any difference, but it fixes a PHPStan error

phpstan-unsupported-baseline.neon
AlexandruGG: @drupol so after testing I realised this happens when passing a callable which takes multiple parameters, both to the ClosureIterator and the Collection object itself. Why? I'm not entirely sure but see the latest commit and the other comment for the changes I did and let me know if you think that works

src/Collection.php
drupol: Sure technically this is valid, it's just confusing according to me, this is why I made the change.

I wish I could understand why PHPStan doesn't like it.

AlexandruGG:
@drupol thank you for the work on this 👍

PR #202, opened 2021-10-09
Refactor/remove static of method

drupol:
This PR:

* [x] Remove Operation interface.
* [x] It breaks backward compatibility
* [x] Update PHPStan and PSalm baselines

WDYT @AlexandruGG ?

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,6 +1,51 @@
 parameters:
 	ignoreErrors:
 		-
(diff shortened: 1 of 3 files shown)


AlexandruGG:
@drupol this is potentially good because it exposes static analysis issues, as evidenced by the new errors reported and the baselines generated for both PHPStan and Psalm.

However, I don't think we should do it all in one go like this and create the baselines. Instead, we should have an issue tracker and gradually move each operation to this new form while **fixing the static analysis issues**. If we keep track somewhere and split the work we can accomplish that a lot faster, and I think it would really benefit the project 😁

Issue #208, opened 2021-10-25
Modify `all` operation to prevent data loss


AlexandruGG:
### Description

It's well-documented both when reading about the [all operation](https://loophp-collection.readthedocs.io/en/latest/pages/api.html#all) and in the section [working with keys and values](https://loophp-collection.readthedocs.io/en/latest/pages/usage.html#working-with-keys-and-values) that when converting a `Collection` to a PHP array, potential data loss can occur. 

This behaviour stems from the flexibility that PHP Generators offer in terms of using multiple times the same key for example. However, in multiple discussions I've had it seemed that for the user this is an "implementation detail", and having to account for this potential data loss can become a cognitive burden.

### Current behaviour 

```PHP
$generator = static function (): Generator {
    yield 0 => 'a';

    yield 1 => 'b';

    yield 0 => 'c';

    yield 1 => 'd';
};

$collection = Collection::fromIterable($generator())
    ->all(); // [0 => 'c', 1 => 'd']

$collection = Collection::fromIterable($generator())
    ->normalize()
    ->all(); // [0 => 'a', 1 => 'b', 2 => 'c', 3 => 'd']
```

### Reasoning

I'm curious to understand why dealing with this behaviour is left to the user, rather than it being handled by the package itself. Most of the times we can safely assume that the user would prefer to keep the data when converting to a PHP array. Many PHP array functions, such as `array_merge`, automatically re-index the keys; this means people are used to this behaviour already.

### Proposal

I would like to propose that the `all` operation automatically applies `normalize` before converting to an array. This removes the need for the user to remember to do this and simplifies the API.

#### Options:
1. Keep backwards compatibility -> in this version we'd introduce a parameter which could be passed to `all` in order to apply the normalization. The existing behaviour would be kept by default. Thus, the signature becomes:
```PHP
public function all(bool $normalize = false): array
```

2. Break backwards compatibility, but still allow old behaviour -> similar to Option 1 but the parameter is `$normalize = true` by default. This ensures that if the user wants the old behaviour they can still have it, but by default the operation will now prevent data loss.

3. Break backwards compatibility, remove old behaviour altogether -> in this version we keep the function signature intact but we always apply `normalize`. This leads to a simpler and easier to use API.

Let me know what you think @drupol :).

AlexandruGG:
> After giving a second though, I prefer option 3.

After looking at this a bit more and testing, I realised that option 3 does not work when using Collection as a map, e.g.

```PHP
[...]

PR #209, opened 2021-10-30
Update `all` operation to prevent data loss

AlexandruGG:
This PR:

* [x] Modifies the `all` operation to apply `normalize` by default. Also allows usage without `normalize` via a bool param
* [x] It breaks backward compatibility
* [x] Has documentation

Fixes https://github.com/loophp/collection/issues/208.

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -12,6 +12,7 @@ $config = require __DIR__ . '/vendor/drupol/php-conventions/config/php73/php_cs_
 $config
     ->getFinder()
     ->ignoreDotFiles(false)
+    ->notPath('src/Contract/Operation/Allable.php')
     ->name(['.php_cs.dist']);
 
 $rules = $config->getRules();
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 
 $config = require __DIR__ . '/vendor/drupol/php-conventions/config/php73/php_cs_fixer.config.php';
 
+// specific file ignore introduced based on discussion: https://github.com/loophp/collection/pull/209#discussion_r739684664
 $config
     ->getFinder()
     ->ignoreDotFiles(false)
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -1283,6 +1283,21 @@ Signature: ``Collection::isEmpty(): bool;``
 .. literalinclude:: code/operations/isEmpty.php
   :language: php
 
+jsonSerialize
+~~~~~~~~~~~~~
+
+Returns the collection items as an array, allowing serialization. Essentially calls ``all(false)``, 
+which means the collection is not normalized by default.
+
+See the section on :ref:`serialization <Serialization>`.
+
+Interface: `JsonSerializable`_
+
+Signature: ``Collection::jsonSerialize(): array;``
+
+.. literalinclude:: code/operations/jsonSerialize.php
+  :language: php
+
 key
 ~~~
 
@@ -2602,6 +2617,7 @@ Signature: ``Collection::zip(iterable ...$iterables): Collection;``
 .. _Doctrine Collections: https://github.com/doctrine/collections
 .. _Generator: https://www.php.net/manual/en/language.generators.overview.php
 .. _Iterator: https://www.php.net/manual/en/class.iterator.php
+.. _JsonSerializable: https://www.php.net/manual/en/class.jsonserializable.php
 .. _symfony/var-dumper: https://packagist.org/packages/symfony/var-dumper
 .. _Traversable: https://www.php.net/manual/en/class.traversable.php
 .. _TypedIterator: https://github.com/loophp/collection/blob/master/src/Iterator/TypedIterator.php
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -70,7 +70,7 @@ Signature: ``Collection::fromGenerator(Generator $generator): Collection;``
     $generator->next();
 
     $collection = Collection::fromGenerator($generator)
-        ->all(); // [2 => 3, 3 => 4, 4 => 5]
+        ->all(); // [0 => 3, 1 => 4, 2 => 5]
 
 fromIterable
 ~~~~~~~~~~~~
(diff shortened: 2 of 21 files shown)


src/Contract/Operation/Allable.php
```diff
@@ -20,7 +20,10 @@ interface Allable
      *
      * @see https://loophp-collection.readthedocs.io/en/stable/pages/api.html#all
      *
-     * @return array<TKey, T>
+     * @param bool $normalize
+     *
+     * @return list<T>|array<TKey, T>
+     * @psalm-return ($normalize is true ? list<T> : array<TKey, T>)
```
AlexandruGG, line 26: Psalm [conditional types](https://psalm.dev/docs/annotating_code/type_syntax/conditional_types/) allow for more accurately describing the return type here

.php-cs-fixer.dist.php
```diff
@@ -12,6 +12,7 @@
 $config
     ->getFinder()
     ->ignoreDotFiles(false)
+    ->notPath('src/Contract/Operation/Allable.php')
```
AlexandruGG, line 15: the issue with php-cs-fixer is that the rule [no_superfluous_phpdoc_tags](https://cs.symfony.com/doc/rules/phpdoc/no_superfluous_phpdoc_tags.html) removes the `@param bool $normalize` phpdoc annotation, which is needed for the conditional return type.

So here we have a choice between:
- deactivating the rule for the whole project
- ignoring that one file

Note that php-cs-fixer does not currently have a mechanism to ignore a rule for a specific file or line, like phpcs has. There are a few issues about this:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3871
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3846

src/Operation/All.php
```diff
@@ -23,16 +23,22 @@ final class All extends AbstractOperation
     /**
      * @pure
      *
-     * @return Closure(Iterator<TKey, T>): array<TKey, T>
+     * @return Closure(bool): Closure(Iterator<TKey, T>): list<T>|array<TKey, T>
```
AlexandruGG, line 26: in this file we cannot use the conditional return type from the interface because we cannot reference the boolean parameter in the same way in many of the closures. So this union return type is the next best thing

AlexandruGG:
@drupol hey, let me know what you think of the comments above and the implementation; if all is good I will continue to complete the outstanding checkboxes

src/Contract/Operation/Allable.php
drupol: So sad we don't have a generic way to do that :(

Besides that, it's fine.

.php-cs-fixer.dist.php
drupol: Oh. That's very unfortunate.

We could either remove that specific rule, or remove the file. What's the best choice?

src/Operation/All.php
drupol: Fair enough.

drupol:
It seems to go in the proper direction, nothing to add here, all good.

.php-cs-fixer.dist.php
AlexandruGG: I think I would prefer to just remove the file for now, it means that we're still getting the benefit of the rule in most of the project. And even with php-cs-fixer ignoring the file we still have phpcs which will enforce style things on it

drupol: I also think this is the best option.

drupol: It's not mandatory because we are using Git, but how about adding a comment inside the PHPCS configuration file pointing to this PR as a reference for later?

drupol (approved):
:100: !

docs/pages/api.rst
```diff
@@ -221,14 +221,15 @@ When used as a ``Collection`` method, operations fall into a few main categories
...
 ~~~
 
-Convert the collection into an array.
+Convert the collection into an array, re-indexing keys using ``Collection::normalize()`` to prevent data loss by default.
 
-.. warning:: This is a lossy operation because PHP array keys cannot be duplicated and must either be ``int`` or ``string``.
-            If you want to ensure no data is lost in the case of duplicate keys, look at the ``Collection::normalize()`` operation.
+.. warning:: An earlier version of this operation did not re-index keys by default, which meant at times data loss could occur.
+        The reason data loss could occur is because PHP array keys cannot be duplicated and must either be ``int`` or ``string``. 
+        The old behaviour can still be achieved by using the operation with the ``$normalize`` parameter as *false*.
```
drupol, line 228: I think we should remove this parameter in the next major version and keep this behavior, WDYT ?

To restore the previous behavior, users can do: `iterator_to_array($collection->getIterator());`

docs/pages/code/operations/all.php
```diff
@@ -30,10 +26,13 @@
...
-// Example 2 -> standalone usage
-$even = static fn (int $value): bool => $value % 2 === 0;
+$collection = Collection::fromIterable($generator());
+print_r($collection->all()); // ['a', 'b', 'c', 'd']
+
+// Example 2 -> usage with "map" collection
+$collection = Collection::fromIterable(['foo' => 1, 'bar' => 2]);
 
-$piped = Pipe::of()(Filter::of()($even), All::of())(new ArrayIterator([1, 2, 3, 4]));
-print_r($piped); // [2, 4]
```
AlexandruGG, line 39: this standalone usage is actually not valid. In order to use `Pipe` every operation needs to return an `Iterator`, but `All` returns an `array`.

This was missed when introducing the `All` operation because there isn't any enforcement of the return type at the [Operation](https://github.com/loophp/collection/blob/master/src/Contract/Operation.php) interface level. It's not easy to implement that however because you could return multiple closures and only the last one needs to return an `Iterator`.

I will look to make a separate PR after this to update the `All` operation to return a `Generator` though, and move the conversion to array back into the Collection method

drupol: Indeed, let's fix that in the next PR.

src/Collection.php
```diff
@@ -624,7 +624,7 @@ public function isEmpty(): bool
      */
     public function jsonSerialize(): array
     {
-        return $this->all();
+        return $this->all(false);
```
AlexandruGG, line 627: we cannot normalize by default in `jsonSerialize` because that would break usage with a "map" collection, e.g.:

```PHP
// Correct -> using `all(false)`
json_encode(Collection::fromIterable(['foo' => 1, 'bar' => 2])); // '{"foo": 1, "bar": 2}'

// Incorrect -> using the new `all()`
json_encode(Collection::fromIterable(['foo' => 1, 'bar' => 2])); // '[1, 2]'
```

@drupol also I realised there's no mention in the documentation that Collection can be serialized. I am thinking to write about it and mention this behaviour as well. What do you think is the best place for it? I would say we could mention it in 2 places:
- `usage.rst` -> have a short "Serialization" section
- `api.rst` -> mentioning the `jsonSerialize` method

drupol: Oh! Nice catch.

I would do it in both files - as serialization is a feature AND part of the API.

docs/pages/api.rst
AlexandruGG: it's a bit trickier because of this: https://github.com/loophp/collection/issues/208#issuecomment-955522900

AlexandruGG:
@drupol hey, I'd say this is ready for final review. Let me know if there's anything else we should add/modify in here

drupol (approved):
LGTM, ready to be merged!

drupol:
Youhou! Thanks!

AlexandruGG:
> Youhou! Thanks!

Thank you as well!

PR #212, opened 2021-11-01
Fix/update PHPStan

### LINKED ISSUE #210: PHPStan 1.0 upgrade
PHPStan 1.0 has been released today.

We have to make sure that PHPStan is still running fine on `loophp/collection` by either fixing the new discovered issues and/or if not applicable by adding them to the baseline files.

We should also make sure it run with the new level 9 that has been introduced in it.

To run phpstan:

```shell
./vendor/bin/phpstan analyze --level 9 src/ tests/static-analysis/
```

drupol:
This PR:

* [x] Fix PHPStan 1.0 new issues (_because of updated level 8_)
* [x] Update baselines and remove obsolete PHPStan issues thanks to new features in PHPStan 1.0
* [x] Update baselines and append new errors detected by PHPStan 1.0
* [x] Fix PHPStan to level 8
* [x] Find why PHPStan fails on PHP 7.4 and not on 8 - Fixed by disabling `reportUnmatchedIgnoredErrors`

Fixes #210

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -6,26 +6,6 @@ parameters:
 			path: src/Operation/Combinate.php
 
 		-
-			message: "#^Method loophp\\\\collection\\\\Operation\\\\Group\\:\\:__invoke\\(\\) should return Closure\\(Iterator\\<TKey, T\\>\\)\\: Generator\\<int, array\\<int, T\\>, mixed, mixed\\> but returns Closure\\(Iterator\\)\\: Generator\\<int, array\\<int, mixed\\>&nonEmpty, mixed, void\\>\\.$#"
-			count: 1
-			path: src/Operation/Group.php
-
-		-
-			message: "#^Strict comparison using \\=\\=\\= between array\\(\\) and array\\<int, T\\>&nonEmpty will always evaluate to false\\.$#"
-			count: 1
-			path: src/Operation/Permutate.php
-
-		-
-			message: "#^Anonymous function should return EmptyIterator but return statement is missing\\.$#"
-			count: 1
-			path: src/Operation/Times.php
-
-		-
-			message: "#^Anonymous function should have native return typehint \"\\?T\"\\.$#"
-			count: 1
-			path: src/Operation/Unwindow.php
-
-		-
 			message: "#^Foreach overwrites \\$value with its value variable\\.$#"
 			count: 1
 			path: src/Operation/Unwindow.php
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,12 +1,292 @@
 parameters:
 	ignoreErrors:
 		-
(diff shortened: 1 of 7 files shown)


AlexandruGG:
Hey, for this:
> Find why PHPStan fails on PHP 7.4 and not on 8 - Maybe we should disable tests on PHP 7.4 ?

Looks like all the failures are related to ignored errors which are actually not found in the code on 7.4. To silence those errors we could just use this config option `reportUnmatchedIgnoredErrors`: https://phpstan.org/user-guide/ignoring-errors#reporting-unused-ignores

phpstan-baseline.neon
```diff
@@ -1,32 +1,487 @@
...
+		-
+			message: "#^Method loophp\\\\collection\\\\Collection\\:\\:empty\\(\\) should return loophp\\\\collection\\\\Collection\\<NewTKey, NewT\\> but returns loophp\\\\collection\\\\Collection\\<\\(int&NewTKey\\)\\|\\(NewTKey&string\\), NewT\\>\\.$#"
+			count: 1
+			path: src/Collection.php
+
+		-
+			message: "#^Parameter \\#1 \\$callable of class loophp\\\\collection\\\\Iterator\\\\ClosureIterator constructor expects callable\\(\\.\\.\\.mixed\\)\\: iterable, Closure\\(Generator\\)\\: Generator\\<mixed, mixed, mixed, void\\> given\\.$#"
+			count: 1
+			path: src/Collection.php
+
```
AlexandruGG, line 12: @drupol I'm a bit concerned by the high number of new errors this is adding to the baseline. This is likely due to PHPStan's new level 9, which is quite brutal, similar to Psalm's level 1 which disallows mixed (which we are not using - this this project uses the default Psalm level 2).

This PR is currently trying to do 2 things at the same time, and I think we should separate them:
- updating to PHPStan 1.0 -> we should be able to do this and fix any errors as long as there aren't too many

- updating to PHPStan level 9 -> I don't think we can do this yet, it might take quite a bit of work to get the static analysis in place properly for this. For now I think we should limit the level to 8 in the [grumphp](https://github.com/drupol/php-conventions/blob/master/config/psr12/grumphp.yml#L88) config for _this project only_

drupol: Right, I had this in mind as well at the beginning, then I skipped it.

I will update the PR accordingly in the evening.

drupol: Job done!

phpstan-unsupported-baseline.neon
```diff
@@ -4,18 +4,3 @@ parameters:
...
-
-		-
-			message: "#^PHPDoc tag @template T for interface loophp\\\\collection\\\\Contract\\\\Operation\\\\Unpackable with bound type array\\<int, NewT\\|NewTKey\\> is not supported\\.$#"
-			count: 1
-			path: src/Contract/Operation/Unpackable.php
-
-		-
-			message: "#^PHPDoc tag @template T for class loophp\\\\collection\\\\Operation\\\\Unpack with bound type array\\<int, NewT\\|NewTKey\\> is not supported\\.$#"
-			count: 1
-			path: src/Operation/Unpack.php
```
AlexandruGG, line 21: something good came out of this 😄

AlexandruGG (approved):
We'll need to see how we can fix the new baseline errors, I'll probably look into it when I get the time

phpstan-unsupported-baseline.neon
drupol: Definitely ! :)

PR #215, opened 2021-11-07
Fix `All` operation for standalone usage

AlexandruGG:
This PR:

* [x] Fixes usage of the `All` as standalone with `Pipe`
* [x] Has documentation

Follows https://github.com/loophp/collection/pull/209#discussion_r743463086. Every operation needs to return an `Iterator` in order to be pipeable.


drupol (approved):
Just tested and reviewed, all good.

drupol:
Thanks again :)

PR #216, opened 2021-11-09
[PHPStan Baseline] Fix a few errors

AlexandruGG:
This PR:

* [x] Fixes a few of the PHPStan baseline errors.

Follows #212

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -354,16 +354,13 @@ final class Collection implements CollectionInterface
     }
 
     /**
-     * Create a new instance with no items.
+     * @pure
      *
-     * @template NewTKey
-     * @template NewT
-     *
-     * @return self<NewTKey, NewT>
+     * @return self<mixed, mixed>
      */
     public static function empty(): CollectionInterface
     {
-        /** @var array<NewTKey, NewT> $emptyArray */
+        /** @var array<mixed, mixed> $emptyArray */
         $emptyArray = [];
 
         return self::fromIterable($emptyArray);
--- a/src/Contract/Collection.php
+++ b/src/Contract/Collection.php
@@ -212,6 +212,7 @@ use loophp\collection\Contract\Operation\Zipable;
  * @template-extends Prependable<TKey, T>
  * @template-extends Productable<TKey, T>
  * @template-extends Randomable<TKey, T>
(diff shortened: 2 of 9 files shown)


src/Collection.php
```diff
@@ -354,16 +354,13 @@ public function duplicate(?callable $comparatorCallback = null, ?callable $acces
...
 
     /**
-     * Create a new instance with no items.
-     *
-     * @template NewTKey
-     * @template NewT
+     * @pure
      *
-     * @return self<NewTKey, NewT>
+     * @return self<mixed, mixed>
```
AlexandruGG, line 359: these template types cannot be inferred from anywhere, so the better option here is to just use `mixed`

src/Contract/Collection.php
```diff
@@ -212,6 +212,7 @@
  * @template-extends Prependable<TKey, T>
  * @template-extends Productable<TKey, T>
  * @template-extends Randomable<TKey, T>
+ * @template-extends Rangeable<TKey, T>
```
AlexandruGG, line 215: somehow this was missing 😄

src/Collection.php
drupol: Fair enough!

src/Contract/Operation/Duplicateable.php
```diff
@@ -9,6 +9,7 @@
 
 namespace loophp\collection\Contract\Operation;
 
+use Closure;
```
AlexandruGG, line 12: found a couple of these where the import was missing but the `Closure` was mentioned in the doc blocks. Because of the missing import PHPStan couldn't resolve the type

src/Contract/Collection.php
drupol: Nice catch :-)

src/Contract/Operation/Duplicateable.php
drupol: Indeed! Nice ones!

drupol (approved):
:+1:

drupol (approved):
Approved!

AlexandruGG:
> Approved!

Thanks 😁 . I'll merge this once the checks have finished

drupol:
Thanks !!

PR #217, opened 2021-11-10
[PHPStan Baseline] Fix Errors Part 2 + Modify `Unfold` Operation

AlexandruGG:
This PR:

* [x] Fixes more PHPStan baseline errors
* [x] Refactors the `Unfold` operation to make it more consistent in usage
* [x] Breaks backwards compatibility
* [x] Has unit tests
* [x] Has static analysis checks
* [x] Has documentation

Follows #216.

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -124,21 +124,7 @@ Signature: ``Collection::range(float $start = 0.0, float $end = INF, float $step
 
 .. code-block:: php
 
-    $fibonacci = static function ($a = 0, $b = 1): array {
-        return [$b, $a + $b];
-    };
-
     $even = Collection::range(0, 20, 2); // [0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20]
-
-Another example
-
-.. code-block:: php
-
-    $even = Collection::unfold(static function ($carry) {return $carry + 2;}, -2);
-    $odd = Collection::unfold(static function ($carry) {return $carry + 2;}, -1);
-    // Is the same as
-    $even = Collection::range(0, \INF, 2);
-    $odd = Collection::range(1, \INF, 2);
 
 times
 ~~~~~
@@ -158,34 +144,14 @@ unfold
 
 Create a collection by yielding from a callback with an initial value.
 
-.. warning:: The callback return values are reused as callback arguments at the next callback call.
+.. warning:: The callback needs to return a list of values which will be reused as callback arguments 
+            on the next callback call. Therefore, the returned list should contain values of the same type
+            as the parameters for the callback function.
 
 Signature: ``Collection::unfold(callable $callback, ...$parameters): Collection;``
 
-.. code-block:: php
-
-    // A list of Naturals from 1 to Infinity.
-    Collection::unfold(fn($n) => $n + 1, 1)
-        ->normalize();
-
-.. code-block:: php
-
-    $fibonacci = static function ($a = 0, $b = 1): array {
-        return [$b, $a + $b];
-    };
-
-    Collection::unfold($fibonacci)
-        ->limit(10); // [[0, 1], [1, 1], [1, 2], [2, 3], [3, 5], [5, 8], [8, 13], [13, 21], [21, 34], [34, 55]]
-
-Another example
-
-.. code-block:: php
-
-    $even = Collection::unfold(static function (int $carry): int {return $carry + 2;}, -2);
-    $odd = Collection::unfold(static function (int $carry): int {return $carry + 2;}, -1);
-    // Is the same as
-    $even = Collection::range(0, \INF, 2);
-    $odd = Collection::range(1, \INF, 2);
+.. literalinclude:: code/operations/unfold.php
+  :language: php
 
 Methods (operations)
 --------------------
@@ -2305,17 +2271,14 @@ Signature: ``Collection::until(callable ...$callbacks): Collection;``
 .. code-block:: php
 
     // The Collatz conjecture (https://en.wikipedia.org/wiki/Collatz_conjecture)
-    $collatz = static function (int $value): int
-    {
-        return 0 === $value % 2 ?
-            $value / 2:
-            $value * 3 + 1;
-    };
+    $collatz = static fn (int $value): array => 0 === $value % 2 
+            ? [$value / 2]
+            : [$value * 3 + 1];
 
     $collection = Collection::unfold($collatz, 10)
-        ->until(static function ($number): bool {
-            return 1 === $number;
-        });
+        ->unwrap()
+        ->normalize()
+        ->until(static fn ($number): bool => 1 === $number);
 
 unwindow
 ~~~~~~~~
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -2277,7 +2277,6 @@ Signature: ``Collection::until(callable ...$callbacks): Collection;``
 
     $collection = Collection::unfold($collatz, 10)
         ->unwrap()
-        ->normalize()
         ->until(static fn ($number): bool => 1 === $number);
 
 unwindow
--- a/docs/pages/code/collatz-conjecture.php
+++ b/docs/pages/code/collatz-conjecture.php
@@ -18,7 +18,6 @@ $collatz = static fn (int $value): array => 0 === $value % 2
 
 $collection = Collection::unfold($collatz, 10)
     ->unwrap()
-    ->normalize()
-    ->until(static fn ($number): bool => 1 === $number);
+    ->until(static fn (int $number): bool => 1 === $number);
 
 print_r($c->all()); // [25, 76, 38, 19, 58, 29, 88, 44, 22, 11, 34, 17, 52, 26, 13, 40, 20, 10, 5, 16, 8, 4, 2, 1]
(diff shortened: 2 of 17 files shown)


src/Iterator/TypedIterator.php
```diff
@@ -37,10 +38,8 @@ public function __construct(Iterator $iterator, ?callable $getType = null)
              * @param mixed $variable
              */
             static function ($variable): string {
-                $type = gettype($variable);
-
-                if ('object' !== $type) {
-                    return $type;
+                if (!is_object($variable)) {
```
drupol, line 41: #AutoFacepalm

src/Operation/Unfold.php
```diff
@@ -25,19 +25,19 @@ final class Unfold extends AbstractOperation
...
             /**
              * @param T ...$parameters
              *
-             * @return Closure(callable(mixed|T...): (array<TKey, T>)): Closure(): Generator<int, T>
+             * @return Closure(callable(T...): (list<T>)): Closure(): Generator<int, T>
              */
             static fn (...$parameters): Closure =>
                 /**
-                 * @param callable(mixed|T...): (mixed|array<TKey, T>) $callback
+                 * @param callable(T...): (T|list<T>) $callback
```
drupol, line 40: I don't understand this change.

`$callback` is a user callback, so, it might return anything... no?

src/Iterator/TypedIterator.php
AlexandruGG: turns out the previous comparison is not interpreted by PHPStan in the same way that the dedicated function comparison is

drupol: Great finding!

src/Operation/Unfold.php
AlexandruGG: well, yes and no.

It's a bit tricky to type this one properly to be honest, so it's up to us if we want to put any restrictions on it or not.

From what I see, this is the behaviour:
- the callback can return _any value_, let's call it "x"
- in the next call, "x" is passed to the callback as parameter, with the following behaviour:
         - if "x" _is a list_, it's unpacked and each item in the list will be a different parameter for the callback
         - if "x" _is not a list_, then it means it will be the parameter for the callback, and the callback only takes one argument

Examples of each:

```PHP
$fibonacci = static fn ($a = 0, $b = 1): array =>[$b, $a + $b];
Collection::unfold($fibonacci);

$even = static fn (int $carry): int => $carry + 2;
Collection::unfold($even, -2);
```

The change I made here means that we are linking the return type of the callback with the parameters that are passed in, but it does also mean that the parameters should be the same type. So we're saying that you can either return a list of parameters of the same type OR a single value which should be the same type as the callback param type.

I think this is fine for most use cases. However, if we want to allow more flexibility the only other option is to go full mixed:

```PHP
@param callable(mixed ...): mixed $callback
```

drupol: I agree. The return type of the closure must be a list in order to fulfill the operation correction.

Maybe we should then get rid of the `(array)` cast and `array_values` then ? WDYT ?

AlexandruGG: yeah those are not needed I think, I'll test it

src/Operation/Unfold.php
```diff
@@ -11,6 +11,7 @@
 
 use Closure;
 use Generator;
+
```
drupol, line 14: :laughing:

src/Operation/Unfold.php
AlexandruGG: I removed both now and introduced logic to check whether the `$parameters` is a list or a single value. Works the same way as before but makes the code more explicit than the `(array)` cast + `array_values`

AlexandruGG: > Yes it's nice!
> 
> How about being more strict and remove the `is_array()` ?

Oh, are you suggesting to modify the operation to only allow us to return an `array` from the callback?

drupol: Yes, definitely, that's what this callback is supposed to return, a list.

AlexandruGG: Ok, I'll make the needed changes then. It will be a minor backwards compatibility break

drupol: I think it might be better to do it now.

AlexandruGG: yeah I'll do it as part of this PR

AlexandruGG: I've updated the code and unit tests, have a look how the usage changes. If you're happy with this I can then go ahead and update the documentation and add some static analysis checks

src/Operation/Unfold.php
```diff
@@ -25,30 +25,29 @@ final class Unfold extends AbstractOperation
...
                 static fn (callable $callback): Closure =>
                     /**
-                     * @return Generator<int, T>
+                     * @return Generator<int, list<T>>
                      */
                     static function () use ($parameters, $callback): Generator {
                         while (true) {
-                            /** @var T $parameters */
-                            $parameters = $callback(...array_values((array) $parameters));
+                            $parameters = $callback(...$parameters);
```
drupol, line 50: I think we should not create a local variable and `yield` directly the result of the callback.

src/Operation/Unfold.php
drupol: yes definitely :)

src/Operation/Unfold.php
AlexandruGG: that won't work the same way because the Generator uses the variable in the next yield iteration here:
`$callback(...$parameters)` 

If there's no variable then it will continue to use the _initial_ parameters

drupol: #TripleFacepalm ... damn I'm too tired!

AlexandruGG: > #TripleFacepalm ... damn I'm too tired!

Don't worry about it 😄. I tried it as well haha, had the same idea initially

drupol: 😌😓

drupol:
I'll wait for this to be done, then do the release.

Tomorrow we're off in Belgium, I might do this tomorrow.

AlexandruGG:
> I'll wait for this to be done, then do the release, if you need some help, just let me know.
> 
> Tomorrow we're off in Belgium, I might do this tomorrow.

I'm not sure if I'll get this done today but we should definitely include it in the release, so I'll tag you when it's ready 😉

docs/pages/code/collatz-conjecture.php
```diff
@@ -12,17 +12,13 @@
...
-$c = Collection::unfold($collatz, 25)
-    ->until(
-        static function ($number): bool {
-            return 1 === $number;
-        }
-    );
+$collection = Collection::unfold($collatz, 10)
+    ->unwrap()
+    ->normalize()
+    ->until(static fn ($number): bool => 1 === $number);
```
drupol, line 22: We could simplify this as:

```php
$collection = Collection::unfold($collatz, 10)
    ->unwrap()
    ->until(static fn (int $number): bool => 1 === $number);
```

AlexandruGG: right, that works now because `all` normalizes by default

drupol: :+1:

docs/pages/api.rst
```diff
@@ -124,22 +124,8 @@ Signature: ``Collection::range(float $start = 0.0, float $end = INF, float $step
...
-Another example
-
-.. code-block:: php
-
-    $even = Collection::unfold(static function ($carry) {return $carry + 2;}, -2);
-    $odd = Collection::unfold(static function ($carry) {return $carry + 2;}, -1);
-    // Is the same as
-    $even = Collection::range(0, \INF, 2);
-    $odd = Collection::range(1, \INF, 2);
-
```
AlexandruGG, line 142: these were repeated in both the `range` section and the `unfold` section. I kept it for the latter

AlexandruGG:
@drupol this should be ready now :)

spec/loophp/collection/CollectionSpec.php
```diff
@@ -3756,39 +3729,26 @@ public function it_can_unpair(): void
...
             return 0 === $initial % 2 ?
-                $initial / 2 :
-                $initial * 3 + 1;
+                [$initial / 2] :
+                [$initial * 3 + 1];
         };
 
         $this::unfold($collatz, 10)
+            ->unwrap()
+            ->normalize()
```
drupol, line 3740: I think we can get rid of `normalize` here as well.

docs/pages/api.rst
drupol: Fair enough.

docs/pages/api.rst
```diff
@@ -2305,17 +2271,14 @@ Signature: ``Collection::until(callable ...$callbacks): Collection;``
...
+    $collatz = static fn (int $value): array => 0 === $value % 2 
+            ? [$value / 2]
+            : [$value * 3 + 1];
 
     $collection = Collection::unfold($collatz, 10)
-        ->until(static function ($number): bool {
-            return 1 === $number;
-        });
+        ->unwrap()
+        ->normalize()
```
drupol, line 2280: The `normalize` can be removed I think.

spec/loophp/collection/CollectionSpec.php
AlexandruGG: you're right although not without updating the test. I made the update now

drupol (approved):
I'll fix the tests :)

drupol:
All good for you @AlexandruGG ? I can merge it ?

PR #218, opened 2021-11-12
Findable: Template param order matters

AlexandruGG:
This PR:

* [x] Fixes an issue with `Findable`: the order of the type parameters matters -  if `T` is first then static analysers will think that in the `callback` provided the first parameter needs to be the key instead of the value.

Not sure how this didn't cause any failures in the static analysis checks we have :(. But it's causing failures in other projects trying to use release 6.0


AlexandruGG:
@drupol could you make a new release 6.0.1 with this one please?

PR #220, opened 2021-11-17
[PHPStan Baseline] A few small fixes

AlexandruGG:
This PR:

* [x] Fixes a few more issues from the PHPStan baseline

### DIFF
--- a/tests/static-analysis/random.php
+++ b/tests/static-analysis/random.php
@@ -38,27 +38,19 @@ function random_checkStringMap(CollectionInterface $collection): void
 }
 
 $intGenerator = static function (): Generator {
-    while (true) {
-        yield mt_rand();
-    }
+    yield mt_rand();
 };
 
 $boolGenerator = static function (): Generator {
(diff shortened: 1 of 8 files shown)


tests/static-analysis/random.php
```diff
@@ -38,27 +38,19 @@ function random_checkStringMap(CollectionInterface $collection): void
 }
 
 $intGenerator = static function (): Generator {
-    while (true) {
```
AlexandruGG, line 41: for now `while (true)` is a new PHPStan error: https://github.com/phpstan/phpstan/issues/5917.

These ones don't matter for us though, makes no difference for the static analysis checks

drupol: Nice !

drupol (approved):
Feel free to merge this even if the CS is not green, I'll fix it a bit later, I'm definitely burried with work.

AlexandruGG:
@drupol thanks for the CS fixer config change 💯

PR #225, opened 2021-11-29
Fix PHP 8.1 Deprecations

AlexandruGG:
This PR fixes new deprecations in PHP 8.1, found when using Collection in a project running this new version.
Note these can be treated as errors in local development and testing depending on the user's PHP error level config.

`
  Error: During inheritance of Iterator: Uncaught Behat\Testwork\Call\Exception\CallErrorException: 8192: Return type of loophp\collection\Iterator\ProxyIterator::key() should either be compatible with Iterator::key(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in [my-project-path]/vendor/loophp/collection/src/Iterator/ProxyIterator.php line 49 in [my-project-path]/vendor/behat/behat/src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php:90
`

* [x] adds ReturnTypeWillChange attributes to iterator methods

### DIFF
--- a/src/Iterator/ArrayCacheIterator.php
+++ b/src/Iterator/ArrayCacheIterator.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace loophp\collection\Iterator;
 
 use Iterator;
+use ReturnTypeWillChange;
 
 use function array_key_exists;
 
@@ -41,6 +42,7 @@ final class ArrayCacheIterator extends ProxyIterator
     /**
      * @return T
      */
+    #[ReturnTypeWillChange]
     public function current()
     {
         $data = $this->getTupleFromCache($this->key);
@@ -51,6 +53,7 @@ final class ArrayCacheIterator extends ProxyIterator
     /**
      * @return TKey
      */
+    #[ReturnTypeWillChange]
     public function key()
     {
         $data = $this->getTupleFromCache($this->key);
--- a/src/Iterator/ProxyIterator.php
+++ b/src/Iterator/ProxyIterator.php
@@ -11,6 +11,7 @@ namespace loophp\collection\Iterator;
 
 use Iterator;
 use OuterIterator;
+use ReturnTypeWillChange;
 
 /**
  * @internal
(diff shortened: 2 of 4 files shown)


src/Iterator/ArrayCacheIterator.php
```diff
@@ -41,6 +42,7 @@ public function __construct(Iterator $iterator)
     /**
      * @return T
      */
+    #[ReturnTypeWillChange]
     public function current()
```
AlexandruGG, line 46: PHP 8.1 now requires the `mixed` return type: https://www.php.net/manual/en/class.iterator.php.

However, if we want Collection to continue supporting PHP 7.4 we cannot add that as it was only introduced in PHP 8. For now the workaround to silence the deprecation is to add this `#[ReturnTypeWillChange]` attribute. More details here: https://wiki.php.net/rfc/internal_method_return_types

drupol: We could work on version 7, which would be only for php >= 8.1 !

AlexandruGG: Hmm we could but it likely means we would need to continue supporting two versions: one for PHP 7.4 and one for >= PHP 8.1. This is because many projects will not be able to update to >= 8.0 and will continue to be on 7.4 for quite some time.

I think for now this is a good solution until PHP 7.4 reaches end of life in November 2022: https://www.php.net/supported-versions.php. We can stop supporting it in this library at that time

AlexandruGG:
Something weird going on with Mutation testing on 7.4, it's been running for > 50 min. I'll try re-running

AlexandruGG:
@drupol hey, shall we get this merged and released? :)

drupol:
Sure @AlexandruGG , sorry for being a bit unresponsive - buried with work here, feel free to merge minor things like that without my approval !

AlexandruGG:
> Sure @AlexandruGG, sorry for being a bit unresponsive - buried with work here, feel free to merge minor things like that without my approval !

No worries, thank you I will merge it and leave it for you to release when you get the time :)

PR #226, opened 2021-12-11
Fix intersperse operation.

SebLours:
This PR:

* [x] Fix intersperse operation
* [ ] Provide ...
* [ ] Has documentation

I think there is a bug in the intersperse operation with a starting position.\
I change the intersperse spec with the expected behavior.

### DIFF
--- a/spec/loophp/collection/CollectionSpec.php
+++ b/spec/loophp/collection/CollectionSpec.php
@@ -2139,8 +2139,6 @@ class CollectionSpec extends ObjectBehavior
             ->shouldIterateAs($generator());
 
         $generator = static function () {
-            yield 0 => 'foo';
-
             yield 0 => 'A';
 
             yield 1 => 'B';
--- a/src/Operation/Intersperse.php
+++ b/src/Operation/Intersperse.php
@@ -62,8 +62,11 @@ final class Intersperse extends AbstractOperation
                                 );
                             }
 
+                            $current = 0;
+
                             foreach ($iterator as $key => $value) {
-                                if (0 === $startAt++ % $atEvery) {
+                                if ($startAt === $current++) {
+                                    $startAt += $atEvery;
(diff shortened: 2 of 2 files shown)


drupol:
Hi!

Thanks for the patch :)

Could you please explain what this patch fix with an example?

Thanks :)

SebLours:
Hi,

Of course, based on the spec.

Before the fix
```
$this::fromIterable(range('A', 'F'))
    ->intersperse('foo', 2, 2)
    ->shouldIterateAs($generator());
```
returns a collection like
```
yield 0 => 'foo';
yield 0 => 'A';
yield 1 => 'B';
yield 2 => 'foo';
yield 2 => 'C';
yield 3 => 'D';
yield 4 => 'foo';
yield 4 => 'E';
yield 5 => 'F';
```
and I think that with a `startAt` argument set as `2`, it must be
```
yield 0 => 'A';
yield 1 => 'B';
yield 2 => 'foo';
yield 2 => 'C';
yield 3 => 'D';
yield 4 => 'foo';
yield 4 => 'E';
yield 5 => 'F';
```
The first item is not `yield 0 => 'foo';` cause we start at `2`

drupol:
You're right, there is a bug.

I'm going to commit your fix and make some improvements later on.

drupol:
Thanks @SebLours !

drupol:
I updated the operation and it's now using iterators, see commit https://github.com/loophp/collection/commit/a208d4c8fe4877cff108196f481fdd7935513553.

Going to create a bugfix release now.

drupol:
Release published! https://github.com/loophp/collection/releases/tag/6.0.3

Thanks for your contribution and for catching this issue!

Let me know if there is some other issue.

SebLours:
Thanks to you for your hard work and this awesome library!

PR #227, opened 2021-12-12
Tests: Switch to PHPUnit

drupol:
This PR:

* [x] Switch to [PHPUnit](https://www.phpunit.de) framework and get rid of [PHPSpec](http://phpspec.net/).
* [x] Use a contrib plugin for PHPUnit to compare iterables: https://github.com/loophp/phpunit-iterable-assertions 
* [x] Depends on #228 because the PHPUnit plugin needs a `ClosureIterator` so I created [`loophp/iterators`](https://github.com/loophp/iterators) to use it in `loophp/collection` and `loophp/phpunit-iterable-assertions`.
* [x] Add a `phpunit.xml` configuration file
* [x] Use PHP 7.4 and 8.1 on continuous integration 
* [x] Fix issue with infection, see https://github.com/infection/infection/issues/1615
* [x] Investigate why Infection score is different on PHP 7.4 and 8.1.

This is a huge work - I'm testing the best way to implement tests. So far I tested:
- [x] One file per test - One dataprovider per file
- [x] One big file containing all the dataproviders and one single test

I really like having one single basic tests and multiple dataproviders... if you have a better idea, let me know.

Follows https://github.com/loophp/collection/discussions/173

PR #228, opened 2021-12-15
Use `loophp/iterators`

drupol:
This PR:

* [x] Use [`loophp/iterators`](https://github.com/loophp/iterators) for providing `ClosureIterator` and `IterableIterator`.
* [x] Get rid of `ProxyIterator` class.
* [x] Get rid of `ArrayCacheIterator` - replaced with a better implementation in `loophp/iterators`. 
* [x] Update `Pack` operation, use `PackIterableAggregator`.
* [x] Iterators are no more implementing `OuterIterator`, it was not needed anywhere. 

Related to #227


src/Iterator/ArrayCacheIterator.php
```diff
@@ -42,7 +46,6 @@ public function __construct(Iterator $iterator)
     /**
      * @return T
      */
-    #[ReturnTypeWillChange]
```
AlexandruGG, line 45: should this be removed? This will create a deprecation in PHP 8.1: https://github.com/loophp/collection/pull/225

src/Iterator/ArrayCacheIterator.php
```diff
@@ -53,7 +56,6 @@ public function current()
     /**
      * @return TKey
      */
-    #[ReturnTypeWillChange]
```
AlexandruGG, line 56: same here and for the new methods

src/Iterator/StringIterator.php
```diff
@@ -47,4 +50,35 @@ static function (string $input, string $delimiter): Generator {
 
         $this->iterator = new IteratorIterator($callback($data, $delimiter));
     }
+
+    /**
+     * @return string
+     */
+    public function current()
```
AlexandruGG, line 57:
```suggestion

    public function current(): string
```

why not like this? Same for others where the type is known

src/Iterator/ArrayCacheIterator.php
drupol: I removed them because I haven't seen them popping in the continuous integration... could you try on your side?

AlexandruGG: I don't have my new laptop set up yet 😅. But in a previous project running on 8.1 they appear as deprecations when running unit tests using [Symfony PHPUnit Bridge](https://symfony.com/doc/current/components/phpunit_bridge.html), which reports deprecations

drupol: Ok, same here. Let's wait a bit then!

AlexandruGG: This is documented though, and there's no downside to not doing it. See https://www.php.net/manual/en/migration81.incompatible.php#migration81.incompatible.core.type-compatibility-internal.

Every iterator needs to conform to the [Iterator](https://www.php.net/manual/en/class.iterator) interface, which means the return types need to be specified inline on the methods. However, `mixed` as return type is not supported in PHP 7.4, which is why I went with this approach [in this PR](https://github.com/loophp/collection/pull/225)

drupol: Okay, going to add them everywhere.

drupol: Done.

src/Iterator/ArrayCacheIterator.php
drupol: Done.

src/Iterator/StringIterator.php
drupol: Done.

AlexandruGG (approved):
Nice 👍

drupol:
Thanks @AlexandruGG !

PR #231, opened 2022-01-06
Use `new` keywords

drupol:
## Motivation

There are multiple goals behind this PR.

First, is to get rid of `AbstractOperation`. Its sole purpose is to provide the static method `of`, which is just syntactic sugar.
Syntactic sugar is cool, but it's not needed in Collection. We could gain some performance by completely removing them.

Then, by removing it and instantiating operation using `new` could also be a good thing for static analysis and types inference.

## Note

When dealing with methods having the `@pure` annotation, using the keyword `new` to create operation will trigger an `ImpureFunctionCall` in PSalm.

However, when having nested functions in those methods, using the `new` keyword, will not trigger an `ImpureFunctionCall` in PSalm.

This PR:

* [x] Use the `new` keyword instead of using the static method `of()`.
* [x] PSalm score is now 97% :tada:

### DIFF
--- a/src/Operation/Map.php
+++ b/src/Operation/Map.php
@@ -28,7 +28,7 @@ final class Map extends AbstractOperation
      *
      * @template V
      *
-     * @return Closure(callable(T=, TKey=, Iterator<TKey, T>=): V): Closure(Iterator<TKey, T>): Generator<TKey, V>
+     * @return Closure(callable(T=, TKey=, Iterator<TKey, T>=): mixed): Closure(Iterator<TKey, T>): Generator<TKey, mixed>
      */
     public function __invoke(): Closure
     {
--- a/src/Operation/Unpack.php
+++ b/src/Operation/Unpack.php
@@ -21,22 +21,19 @@ use Traversable;
 /**
  * @immutable
  *
- * @template NewTKey
- * @template NewT
- *
  * @template TKey
- * @template T of array{0: NewTKey, 1: NewT}
+ * @template T
  */
 final class Unpack extends AbstractOperation
 {
     /**
      * @pure
      *
-     * @return Closure(Iterator<TKey, T>): Generator<NewTKey, NewT>
+     * @return Closure(Iterator<mixed, mixed>): Generator<TKey, T>
      */
     public function __invoke(): Closure
     {
-        /** @var Closure(Iterator<TKey, T>): Generator<NewTKey, NewT> $pipe */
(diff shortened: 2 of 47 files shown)


src/Operation/Unpack.php
```diff
@@ -21,22 +21,19 @@
...
- * @template T of array{0: NewTKey, 1: NewT}
+ * @template T
  */
 final class Unpack extends AbstractOperation
 {
     /**
      * @pure
      *
-     * @return Closure(Iterator<TKey, T>): Generator<NewTKey, NewT>
+     * @return Closure(Iterator<mixed, mixed>): Generator<TKey, T>
```
AlexandruGG, line 32: this change is not beneficial because it's restricting the types of the keys and values to the original TKEY, T, which is not the case typically with `unpack` in practice.

I guess if we want to get rid of the other template parameters (because they cannot be inferred), we will have to go with `mixed` here, e.g. `Generator<mixed, mixed>`.

Also, any changes made here should also be reflected in the interface Unpackable and in the static analysis checks

AlexandruGG (dismissed):
Nice 👍 . I like the idea of this. 

All looks good except for the `Unpack` changes. Perhaps you could leave those for a separate PR and address the comment?

src/Operation/Unpack.php
drupol: Going to do it in here this evening then ! Thanks for the feedback :)🙏

drupol:
> Nice 👍 . I like the idea of this. 
> 
> All looks good except for the `Unpack` changes. Perhaps you could leave those for a separate PR and address the comment?

Done in this PR.

src/Operation/Map.php
```diff
@@ -28,7 +28,7 @@ final class Map extends AbstractOperation
      *
      * @template V
      *
-     * @return Closure(callable(T=, TKey=, Iterator<TKey, T>=): V): Closure(Iterator<TKey, T>): Generator<TKey, V>
+     * @return Closure(callable(T=, TKey=, Iterator<TKey, T>=): mixed): Closure(Iterator<TKey, T>): Generator<TKey, mixed>
```
AlexandruGG, line 31: if we're changing this here then it should also be reflected in:
- Mapable
- AsyncMap
- AsyncMapable

In general any type changes on the operations should match the interfaces; we don't want to have discrepancies between them

AlexandruGG: same for the scans and others in this PR

drupol: I think it's ready now :)

AlexandruGG (approved):
💯

PR #238, opened 2022-02-24
feat: Add `Averages` operation.

drupol:
This PR:

* [x] Add the `averages` operation
* [x] Has documentation

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -312,6 +312,18 @@ Signature: ``Collection::asyncMapN(callable ...$callbacks): Collection;``
     $collection = Collection::fromIterable(['c' => 3, 'b' => 2, 'a' => 1])
         ->asyncMapN($mapper1, $mapper2); // ['a' => 2, 'b' => 4, 'c' => 6]
 
+averages
+~~~~~~~~
+
+Calculate the average of a list of numbers.
+
+Interface: `Averagesable`_
+
+Signature: ``Collection::averages(): Collection;``
+
+.. literalinclude:: code/operations/averages.php
+  :language: php
+
 cache
 ~~~~~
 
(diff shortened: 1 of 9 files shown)


docs/pages/api.rst
```diff
@@ -312,6 +312,18 @@ Signature: ``Collection::asyncMapN(callable ...$callbacks): Collection;``
     $collection = Collection::fromIterable(['c' => 3, 'b' => 2, 'a' => 1])
         ->asyncMapN($mapper1, $mapper2); // ['a' => 2, 'b' => 4, 'c' => 6]
 
+averages
+~~~~~~~~
+
+Calculate the average of a list of numbers.
```
AlexandruGG, line 318: could you add a few more details to clarify what this means exactly?

I eventually understood after staring at the code and the example for a bit, but it's not immediately clear 😅.

Maybe we could mention that it's using `scanLeft1` under the hood and produces a list in the same fashion.

docs/pages/code/operations/averages.php
```diff
@@ -0,0 +1,20 @@
...
+declare(strict_types=1);
+
+namespace App;
+
+use loophp\collection\Collection;
+
+include __DIR__ . '/../../../../vendor/autoload.php';
+
+$collection = Collection::fromIterable([1, 2, 3, 4, 5])
+    ->averages(); // [1, 1.5, 2, 2.5, 3]
```
AlexandruGG, line 17: here we could give an example of the calculations, like:

```php
// [1, 1.5, 2, 2.5, 3]
// [1/1, (1+2)/2, (1+2+3)/3 ...]
```

src/Contract/Operation/Averagesable.php
```diff
@@ -0,0 +1,28 @@
...
+use loophp\collection\Contract\Collection;
+
+/**
+ * @template TKey
+ * @template T
+ */
+interface Averagesable
+{
+    /**
+     * Compute the averages.
```
AlexandruGG, line 21: the doc block here should match the description in the `api.rst`

src/Operation/Averages.php
```diff
@@ -0,0 +1,43 @@
...
+        return (new Pipe())()(
+            (new Normalize())(),
+            (new ScanLeft1())()(
+                /**
+                 * @param float $acc
+                 * @param float $value
+                 *
+                 * @return float
+                 */
+                static fn ($acc, $value, int $key) => ($acc * $key + $value) / ($key + 1)
```
AlexandruGG, line 39:
```suggestion
                static fn (float $acc, float $value, int $key): float => ($acc * $key + $value) / ($key + 1)
```

why not like this here? 🤔

drupol: Thanks: )

src/Contract/Operation/Averagesable.php
drupol: Done

docs/pages/code/operations/averages.php
drupol: Done.

docs/pages/api.rst
drupol: Done!

AlexandruGG (approved):
Really nice one 🚀

PR #242, opened 2022-03-05
refactor: Update `every` operation.

drupol:
Make it more flexible and fast.

This PR:

* [x] Update `every` operation
* [x] Need to update `TakeWhile` and `DropWhile`
* [ ] Has documentation

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -130,6 +130,7 @@ use loophp\collection\Operation\Window;
 use loophp\collection\Operation\Words;
 use loophp\collection\Operation\Wrap;
 use loophp\collection\Operation\Zip;
+use loophp\collection\Utils\CallbacksArrayReducer;
 use loophp\iterators\ClosureIteratorAggregate;
 use loophp\iterators\IterableIteratorAggregate;
 use loophp\iterators\ResourceIteratorAggregate;
@@ -360,7 +361,8 @@ final class Collection implements CollectionInterface
 
     public function every(callable ...$callbacks): bool
     {
-        return (new Every())()(static fn (): bool => false)(...$callbacks)($this)->current();
+        return (new Every())()(static fn (int $index, $value, $key, iterable $iterable) => CallbacksArrayReducer::or()($callbacks)($value, $key, $iterable))(static fn (bool $r): bool => $r)($this)
+            ->current();
     }
 
     public function explode(...$explodes): CollectionInterface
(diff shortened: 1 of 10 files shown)


src/Collection.php
```diff
@@ -360,7 +361,8 @@ public function equals(iterable $other): bool
 
     public function every(callable ...$callbacks): bool
     {
-        return (new Every())()(static fn (): bool => false)(...$callbacks)($this)->current();
+        return (new Every())()(static fn (int $index, $value, $key, iterable $iterable) => CallbacksArrayReducer::or()($callbacks)($value, $key, $iterable))(static fn (bool $r): bool => $r)($this)
+            ->current();
```
AlexandruGG, line 365: should we not use the `CallbacksArrayReducer` inside the `Every` operation? That would still allow using it with multiple callbacks as a standalone operation

src/Operation/Pluck.php
```diff
@@ -19,6 +19,7 @@
 use function in_array;
 use function is_array;
 use function is_object;
+use function is_scalar;
```
AlexandruGG, line 22: hmm did you add this manually or was it added by php-cs-fixer or something automatically? 🤔

src/Collection.php
drupol: Yes ! This is on my todo list indeed.

src/Operation/Pluck.php
drupol: php-cs-fixer :)

AlexandruGG: cool, I guess it's not set up to import _all native functions_ though, because there are others like `array_shift` and `property_exists` in this file that are not imported

src/Collection.php
drupol: Done.

src/Operation/Pluck.php
drupol: It has been added in here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/6277

drupol: I made a commit in master with these CS changes, so they don't pollute this branch.

AlexandruGG: ah I see, I've typically used that rule with the `all` option rather than only the compiler optimized ones just for consistency, which is why I was surprised 😄

AlexandruGG (approved):

PR #252, opened 2022-05-07
ci: Use `pcov` for coverage.

drupol:
This PR:

* [ ] Fix ...
* [ ] Provide ...
* [ ] Is covered by unit tests
* [ ] Has documentation

Follows #.
Related to #.
Fixes #.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,7 +33,7 @@ jobs:
               uses: shivammathur/setup-php@2.18.1
               with:
                   php-version: ${{ matrix.php-versions }}
-                  coverage: xdebug
+                  coverage: pcov
                   extensions: gd,mbstring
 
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -33,7 +33,8 @@ jobs:
               uses: shivammathur/setup-php@2.18.1
               with:
                   php-version: ${{ matrix.php-versions }}
-                  extensions: gd,mbstring,pcov,xdebug
+                  coverage: xdebug
```
maks-rafalko, line 36: just a qustion: is there any benefit of using `xdebug` for **only** coverage when you don't use branch coverage? IMO `pcov` is much faster

drupol: you're right, I found a way to get it working !

PR #269, opened 2022-11-03
refactor: `unfold` constructor is not variadic anymore

drupol:
This PR:

- [ ] Fix ...
- [ ] Provide ...
- [ ] Has documentation

Follows #. Related to #. Fixes #.

### DIFF
--- a/src/Contract/Operation/Unfoldable.php
+++ b/src/Contract/Operation/Unfoldable.php
@@ -6,6 +6,10 @@ namespace loophp\collection\Contract\Operation;
 
 use loophp\collection\Contract\Collection;
 
+/**
+ * @template TKey
+ * @template T
+ */
 interface Unfoldable
 {
     /**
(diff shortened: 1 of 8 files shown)


src/Contract/Operation/Unfoldable.php
```diff
@@ -6,19 +6,23 @@
...
      *
      * @see https://loophp-collection.readthedocs.io/en/stable/pages/api.html#unfold
      *
-     * @template T
+     * @template U
      *
-     * @param callable(T ...): list<T> $callback
-     * @param T ...$parameters
+     * @param callable((T|U) ...): list<T> $callback
+     * @param array<int, U> $parameters
```
AlexandruGG, line 23: so does this mean that now in `$parameters` we want to be able to pass values of different types? What would be a good use case for that?

drupol: Yes we could potentially do that.

drupol: I gave a second look yesterday and you are perfectly right. I reverted some of the changes.

It's ready again for review.

src/Contract/Operation/Unfoldable.php
```diff
@@ -6,19 +6,23 @@
...
+ */
 interface Unfoldable
 {
     /**
      * Create a collection by yielding from a callback with an initial value.
      *
      * @see https://loophp-collection.readthedocs.io/en/stable/pages/api.html#unfold
      *
-     * @template T
+     * @template U of T
```
AlexandruGG, line 20: I'm not sure why this is needed - doesn't it work if we leave it like before, just `T` ? Either way I guess it's the same thing

AlexandruGG (approved):

src/Contract/Operation/Unfoldable.php
drupol: This is a static function, so, `U` must be defined locally at the method level. And `U` is going to be made of `T`.

PR #287, opened 2022-12-26
Psalm 5.4 upgrade

drupol:
This PR:

- [x] Fix issues introduced by Psalm 5.4
- [x] Needs PSalm > 5.4.0 to work (_use dev version in the meantime_)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,8 @@
     ],
     "require": {
         "php": ">= 8",
-        "loophp/iterators": "^2.3.1"
+        "loophp/iterators": "^2.3.1",
+        "vimeo/psalm": "dev-master#f723e08221c5028612d928241c84d9cac7fe6d86"
     },
     "require-dev": {
         "amphp/parallel-functions": "^1",
--- a/src/Operation/Wrap.php
+++ b/src/Operation/Wrap.php
@@ -16,20 +16,20 @@ use Generator;
 final class Wrap extends AbstractOperation
 {
     /**
-     * @return Closure(iterable<TKey, T>): Generator<int, array<array-key, T>>
+     * @return Closure(iterable<TKey, T>): Generator<int, array<TKey, T>>
      */
     public function __invoke(): Closure
     {
-        /** @var Closure(iterable<TKey, T>): Generator<int, array<array-key, T>> $pipe */
+        /** @var Closure(iterable<TKey, T>): Generator<int, array<TKey, T>> $pipe */
         $pipe = (new Pipe())()(
             (new Map())()(
                 /**
                  * @param T $value
-                 * @param array-key $key
+                 * @param TKey $key
                  *
-                 * @return array<array-key, T>
+                 * @return array<TKey, T>
                  */
-                static fn (mixed $value, int|string $key): array => [$key => $value]
+                static fn (mixed $value, mixed $key): array => [$key => $value]
             ),
             (new Normalize())()
         );
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -26,6 +26,11 @@ parameters:
 			path: src/Collection.php
 
 		-
+			message: "#^Method loophp\\\\collection\\\\Collection\\:\\:squash\\(\\) should return loophp\\\\collection\\\\Contract\\\\Collection\\<TKey, T\\> but returns loophp\\\\collection\\\\Contract\\\\Collection\\<int, array\\{TKey, T\\}\\>\\.$#"
+			count: 1
+			path: src/Collection.php
+
+		-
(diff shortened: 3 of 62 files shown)


drupol:
This PR needs to be merged once the upcoming version of PSalm is out (5.4.1).

composer.json
```diff
@@ -37,7 +37,8 @@
     ],
     "require": {
         "php": ">= 8",
-        "loophp/iterators": "^2.3.1"
+        "loophp/iterators": "^2.3.1",
+        "vimeo/psalm": "dev-master#3560ef771743048302599aea151e206a50f4a744"
```
AlexandruGG, line 41: Psalm 5.4 is out now, is this still needed? 

https://github.com/vimeo/psalm/releases/tag/5.4.0

src/Contract/Operation/Unfoldable.php
```diff
@@ -14,11 +14,12 @@ interface Unfoldable
      * @see https://loophp-collection.readthedocs.io/en/stable/pages/api.html#unfold
      *
      * @template T
+     * @template TKey
```
AlexandruGG, line 17: `TKey` is unused here, did you mean to change the return type on line 22 to `Collection<TKey, iterable<array-key, T>>`?

src/Contract/Operation/Unwrapable.php
```diff
@@ -18,7 +18,7 @@ interface Unwrapable
      *
      * @see https://loophp-collection.readthedocs.io/en/stable/pages/api.html#unwrap
      *
-     * @return Collection<mixed, mixed>
+     * @return Collection<TKey, T>
```
AlexandruGG, line 21: not too sure about this change -> it's not a given that the elements in the unwrapped collection will be all of the same type 🤔 

```PHP
Collection::fromIterable([['a'], [1], [true]])
    ->unwrap(); // ['a', 1, true]
```

unless static analysis is able to interpret the template `T` as the union `string|int|bool` in this case. I'm not sure but you could test this with a static analysis check

tests/static-analysis/unwrap.php
```diff
@@ -8,27 +8,71 @@
...
+ *
+ * @phpstan-param CollectionInterface<int, int> $collection
  */
-function unwrap_checkList(CollectionInterface $collection): void
+function unwrap_checkList1(CollectionInterface $collection): void
 {
 }
 /**
- * @param CollectionInterface<string, string> $collection
+ * @psalm-param CollectionInterface<int<0, 2>, 2|3|list{1}> $collection
```
AlexandruGG, line 19: this is not good, and it shows that the new type hints have made `unwrap` too restrictive and incorrect in this case

```PHP
Collection::fromIterable([[1], 2, 3])->unwrap() 
// will result in Collection(1, 2, 3) aka `Collection<int, int>`, but the type hints think the resulting types are the same 
// as were passed in the original collection
```

src/Contract/Operation/Flattenable.php
```diff
@@ -19,7 +19,7 @@ interface Flattenable
      *
      * @see https://loophp-collection.readthedocs.io/en/stable/pages/api.html#flatten
      *
-     * @return Collection<mixed, mixed>
+     * @return Collection<TKey, T>
```
AlexandruGG, line 22: same issue here as with the `unwrap`

src/Operation/Wrap.php
```diff
@@ -16,25 +16,20 @@
...
-            (new Normalize())()
-        );
-
-        // Point free style.
-        return $pipe;
+        return
+        /**
+         * @param iterable<TKey, T> $iterable
+         *
+         * @return Generator<int, non-empty-array<TKey, T>>
```
AlexandruGG, line 27: this will not always be a `non-empty-array` I think? What if the `iterable` is empty?

AlexandruGG (commented):
I think we need to be careful not to make things way too restrictive -> I see lots of changes needed in the static analysis checks - these are all potential changes users would need to make in their own codebases.

If some of changes are unavoidable simply due to Psalm 5.4 then we can perhaps use a baseline and introduce them more gradually? Would be easier if we would only do it for a set of operations at a time so we can carefully think about how the change affects their usage.

composer.json
drupol: We need the next version, 5.4.1 ;)

drupol:
@AlexandruGG All good for you now?

AlexandruGG:
> @AlexandruGG All good for you now?
> 

I'll give this another review tomorrow 😁

AlexandruGG:
> @AlexandruGG All good for you now?

Hey @drupol, I'm not sure what's changed because still see the main issues I commented about are still present. I'm talking about stuff like Unwrap, Flatten, and generally making things too restrictive with respect to static analysis.

drupol:
Right, I removed the changes from `flatten` and `unwrap`.

drupol:
@AlexandruGG Good for you?

AlexandruGG (approved):
Thanks, I know this took lots of work!

drupol:
Thanks! Indeed it was painful !

PR #288, opened 2022-12-28
Fix `reducing` operations

drupol:
This PR:

- [x] Fix `reduce`, `reduction`, `scanLeft`, `scanRight`
- [x] Has documentation

Note, the psalm issues are related to the upgrade to 5.4, they will be fixed in a follow up PR which is almost done.

Follows a discussion on Mastodon with @azjezz at https://mathstodon.xyz/@Pol/109588243184619262

### DIFF
--- a/tests/static-analysis/reduce.php
+++ b/tests/static-analysis/reduce.php
@@ -18,15 +18,25 @@ function reduce_takeString(string $string): void
 {
 }
 
+function reduce_takeStringOrNull(?string $string): void
+{
+}
+
+function reduce_takeIntOrBool(int|bool $v): void
+{
+}
+
 $sumNullable = static fn (?int $carry, int $value): int => null === $carry ? $value : $carry + $value;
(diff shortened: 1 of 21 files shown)


tests/static-analysis/reduce.php
```diff
@@ -6,15 +6,19 @@
 
 use loophp\collection\Collection;
 
-function reduce_takeInt(int $int): void
+function reduce_takeInt(?int $int): void
```
AlexandruGG, line 9: hmm seems like this one is now the same as `reduce_takeIntOrNull` so perhaps not needed?

AlexandruGG (dismissed):
I saw the loophp/iterators changes as well, thanks for this!

tests/static-analysis/reduce.php
drupol: Indeed! I will update the pr asap

PR #306, opened 2023-09-01
feat: add `countIn` operation

drupol:
This PR:

- [x] Provide a new operation `countIn` inspired from https://github.com/fiiSoft/fiisoft-jackdaw-stream/ project
- [x] Has documentation

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -554,7 +554,7 @@ Signature: ``Collection::count(): int;``
 countIn
 ~~~~~~~
 
-This operation requires a reference to a parameter that will contains the amount
+This operation requires a reference to a parameter that will contain the amount
 of items in the collection.
 
 Interface: `CountInAble`_
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -557,6 +557,10 @@ countIn
 This operation requires a reference to a parameter that will contain the amount
 of items in the collection.
 
+The difference with the `count` operation is that the `count` operation will
+return the amount of items in the collection and the `countIn` operation will
+yield over the collection itself while updating the counter variable.
+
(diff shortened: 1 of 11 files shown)


docs/pages/code/operations/countIn.php
```diff
@@ -0,0 +1,20 @@
...
+$collection = Collection::fromString('The quick brown fox jumps over the lazy dog')
+    ->countIn($lettersCounter)
+    ->words()
+    ->countIn($wordsCounter)
+    ->map(
+        static function (string $word, int $k) use ($wordsCounter): string {
+            return sprintf('[%s/%s]: %s', $k + 1, $wordsCounter, $word);
+        }
+    )
+    ->all();
```
AlexandruGG, line 20: can we add a comment to show what this will contain in the end?

docs/pages/api.rst
```diff
@@ -551,6 +551,19 @@ Signature: ``Collection::count(): int;``
     $collection = Collection::fromIterable(range('a', 'c'))
         ->count(); // 3
 
+countIn
+~~~~~~~
+
+This operation requires a reference to a parameter that will contain the amount
+of items in the collection.
```
AlexandruGG, line 558: might be nice to add a tip to show when this would be useful vs just using `count`

docs/pages/code/operations/countIn.php
drupol: Fixed

docs/pages/api.rst
drupol: Fixed

AlexandruGG (dismissed):
Interesting addition!

drupol:
Thanks for the review :)

PR #314, opened 2023-11-11
feat: add `plus` operation

### LINKED ISSUE #313: Plus operation RFC
## PHP plus operator

I'm wondering if there's same feature as php plus operator:

```php
public function testPlusOperation(): void
{
    $initialArray =[
        'id' => ['A', 'B'],
        'id2' => ['C'],
    ];

    $resultArray = $initialArray + ['id' => [], 'id2' => [], 'id3' => []];

    self::assertSame([
        'id' => ['A', 'B'],
        'id2' => ['C'],
        'id3' => [],
    ], $resultArray);
}
```

As of me, personally I often use plus operation on arrays and it would make sense to implement it within a collection.

## How it'd look like with collection

```php
public function testCollectionPlusOperation(): void
{
    $initialCollection = Collection::fromIterable([
        ['id' => 'A'],
        ['id' => 'B'],
        ['id2' => 'C'],
    ]);

    $result = $initialCollection
        ->unwrap()
        ->groupBy(static fn ($value, $key) => $key)
        ->plus(['id' => [], 'id2' => [], 'id3' => []])
        ->all(false)
    ;

    self::assertSame([
        'id' => ['A', 'B'],
        'id2' => ['C'],
        'id3' => [],
    ], $result);
}
```

## Comparison with `prepend`

With current implementation seemingly the same behavior may be achieved using prepend:

```php
$result = $initialCollection
    ->unwrap()
    ->groupBy(static fn ($value, $key) => $key)
    ->prepend(...['id' => [], 'id2' => [], 'id3' => []])
    ->all(false)
;

self::assertSame([
    'id' => ['A', 'B'],
    'id2' => ['C'],
    'id3' => [],
], $result);
```

However it is not explicit solution, since the collection maintains all the keys until `all()` is called. 

```php
public function testPrependedKeysAreStillPresent(): void
{
    $initialCollection = Collection::fromIterable([
        ['id' => 'A'],
        ['id' => 'B'],
        ['id2' => 'C'],
    ]);

    $result = $initialCollection
        ->unwrap()
        ->groupBy(static fn ($value, $key) => $key)
        ->prepend(...['id' => [], 'id2' => [], 'id3' => []])
        ->pack()
        ->all()
    ;

    self::assertSame([
        ['id', []],
        ['id2', []],
        ['id3', []],
        ['id', ['A', 'B']],
        ['id2', ['C']],
    ], $result);
}
```

Therefore, when using `prepend`, we must convert collection into array before we can safely use the result, lest same instance iteration would yield `id` and `id2` keys twice and `id3` key before the rest.

Please, let me know what you think about this `plus()` proposal.

drupol:
This should fix #313

This PR:

- [ ] Fix ...
- [ ] Provide ...
- [ ] Has documentation

Follows #. Related to #. Fixes #.

### DIFF
--- a/src/Operation/Plus.php
+++ b/src/Operation/Plus.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace loophp\collection\Operation;
+
+use Closure;
+use Generator;
+use loophp\iterators\ConcatIterableAggregate;
+
+/**
+ * @immutable
+ *
+ * @template TKey
+ * @template T
+ */
+final class Plus extends AbstractOperation
+{
+    /**
+     * @template UKey
+     * @template U
+     *
+     * @return Closure(iterable<UKey, U>): Closure(iterable<TKey, T>): iterable<int|TKey|UKey, T|U>
+     */
+    public function __invoke(): Closure
+    {
+        $comparatorCallback =
+            /**
+             * @param T $left
+             *
+             * @return Closure(T): bool
+             */
+            static fn (mixed $left): Closure =>
+                /**
+                 * @param T $right
+                 */
+                static fn (mixed $right): bool => $left === $right;
+
+        return
+            /**
+             * @param iterable<UKey, U> $items
+             *
+             * @return Closure(iterable<TKey, T>): iterable<int|TKey|UKey, T|U>
+             */
+            static fn (iterable $items): Closure =>
+                /**
+                 * @param iterable<TKey, T> $iterable
+                 *
+                 * @return Generator<int|TKey|UKey, T|U>
+                 */
+                static fn (iterable $iterable): Generator => yield from (new Distinct())()($comparatorCallback)(static fn (mixed $value, mixed $key): mixed => $key)(new ConcatIterableAggregate([$iterable, $items]));
+    }
+}
--- a/src/Operation/Plus.php
+++ b/src/Operation/Plus.php
@@ -6,7 +6,7 @@ namespace loophp\collection\Operation;
 
 use Closure;
 use Generator;
-use loophp\iterators\ConcatIterableAggregate;
+use loophp\collection\Collection;
 
 /**
  * @immutable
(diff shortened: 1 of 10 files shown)


src/Operation/Plus.php
```diff
@@ -0,0 +1,53 @@
...
+             *
+             * @return Closure(iterable<TKey, T>): iterable<int|TKey|UKey, T|U>
+             */
+            static fn (iterable $items): Closure =>
+                /**
+                 * @param iterable<TKey, T> $iterable
+                 *
+                 * @return Generator<int|TKey|UKey, T|U>
+                 */
+                static fn (iterable $iterable): Generator => yield from (new Distinct())()($comparatorCallback)(static fn (mixed $value, mixed $key): mixed => $key)(new ConcatIterableAggregate([$iterable, $items]));
```
rela589n, line 51: @drupol , does it make any sense to remove duplicated indexes in the existing collection? I assume it would be more apparent if these were kept. Consider following test (fails with current implementation):

```php
public function testPlusDoesNotRemoveExistingDuplicatedIndices(): void
{
    $collection = Collection::fromIterable([
        ['id' => 1],
        ['id' => 2],
    ]);

    $result = $collection
        ->unwrap()
        ->plus(['id' => 3, 'id2' => 4])
        ->wrap()
        ->all(false)
    ;

    self::assertSame([
        ['id' => 1],
        ['id' => 2],
        ['id2' => 4],
    ], $result);
}
```

Moreover, it would be really pity if `plus([])` would drop all key duplicates. The following test fails as well:

```php
public function testPlusEmptyArrayDoesNotModifyCollection(): void
{
    $collection = Collection::fromIterable([
        ['id' => 1],
        ['id' => 2],
    ]);

    $result = $collection
        ->unwrap()
        ->plus([])
        ->wrap()
        ->all(false)
    ;

    self::assertSame([
        ['id' => 1],
        ['id' => 2],
    ], $result);
}
```

drupol: Very nice catch and test case. I will see how we can improve the implementation. Feel free to come up with ideas.

drupol: I updated the tests and the implementation, this is a bit uglier, but AFAIK there's no other alternative.

drupol:
@rela589n Are you willing to contribute to the documentation for this operation?

drupol:
No more feedback from the reporter, closing this PR.

PR #317, opened 2023-11-16
feat: add `entropy` operation

drupol:
This PR:

- [x] Provide the `entropy` operation
- [x] Has documentation

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -723,6 +723,27 @@ Signature: ``Collection::duplicate(?callable $comparatorCallback = null, ?callab
 .. literalinclude:: code/operations/duplicate.php
   :language: php
 
+entropy
+~~~~~~~
+
+This method extends the library's functionality by allowing users to calculate
+the normalized Shannon entropy of a collection.
+
+The implementation provides the normalized version of Shannon entropy.
+Normalization ensures that the entropy value is scaled between `0` and `1`,
+making it easier to compare entropy across different collections, irrespective
+of their size or the diversity of elements they contain.
+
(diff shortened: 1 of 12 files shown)


docs/pages/api.rst
```diff
@@ -723,6 +723,27 @@ Signature: ``Collection::duplicate(?callable $comparatorCallback = null, ?callab
...
+This method extends the library's functionality by allowing users to calculate
+the normalized Shannon entropy of a collection.
+
+The implementation provides the normalized version of Shannon entropy.
+Normalization ensures that the entropy value is scaled between `0` and `1`,
+making it easier to compare entropy across different collections, irrespective
+of their size or the diversity of elements they contain.
+
+An entropy of `0` indicates no diversity (all elements are the same), while an
+entropy of `1` signifies maximum diversity.
```
AlexandruGG, line 738: would be useful to clarify what the resulting collection means, because just by reading this I would assume that the operation would return a single value. But actually it results in a list of values.

src/Contract/Operation/Entropyable.php
```diff
@@ -0,0 +1,27 @@
...
+ * @template TKey
+ * @template T
+ */
+interface Entropyable
+{
+    /**
+     * Calculate the normalized Shannon entropy for each collection items.
+     *
+     * If you're looking for one single result, you must get the last item using
+     * `last` operation.
```
AlexandruGG, line 19: let's make the doc block description here match the api docs descriptions like the other operations

AlexandruGG (commented):
Interesting operation idea!

AlexandruGG (approved):
Nice! 🚀

Issue #264, opened 2022-09-08
Issue with cache and fromCallable


ktrzos:
Hey 👋. I have this code:

```php
$test = Collection::fromCallable(static function () {
    yield 100 => 'a';
    yield 200 => 'b';
    yield 300 => 'c';
    yield 400 => 'd';
})->cache();

$result1 = $test->get(100)->current();
$result2 = $test->get(200)->current();
$result3 = $test->get(300)->current();
$result4 = $test->get(400)->current();
```

With each run, the result is:

![image](https://user-images.githubusercontent.com/14212721/189178364-c5e29448-3162-4789-95b0-e1a784950069.png)

Using the latest loophp/collection (6.0.3) and PHP 8.1. Is it a bug, or am I doing something wrong 😉?

drupol:
Hi !

Interesting, let me check this out locally with version 6, `master` and I'll get back to you.

ktrzos:
Thank you for VERY fast response @drupol 🍻

drupol:
You're welcome.

I cannot reproduce the issue with the `master` branch, everything seems to be ok there.

Going to investigate the issue on version `^6`...

drupol:
This is the commit who fixed the issue a69acec

ktrzos:
Thanks! I suppose it will be soon pushed on the next library version?

drupol:
I'm still investigating which lines fixed the issue. I need to provide a fix and release a version 6.0.4 with the fix only.

ktrzos:
Ok, no worries. I will wait for it then 🙂. Good luck 🤞!

drupol:
Ok, got the reason.

That commit is actually one of the most important commit. This is mostly the reason why there will be a major version bump.

The sad news is that I'm afraid this won't be fixable in v6 :( it would require to rewrite almost every operations.

The good news is that now I need to hurry and release version 7. I took a bit of delay to finish it. There's still #256 to finish and we're good to go.

ktrzos:
I see. It is ok. I managed to make a workaround on my side by casting the collection to an array (unfortunately), and getting the value by a key at that point. Keep up the good work, and can't wait for v7 🙂.

ktrzos:
And, of course, thank you for your help and fast reaction once again 👍️.

drupol:
I just added your tests in the codebase to make sure it doesn't happen again in v7 :)

Thanks for reporting it !

rela589n:
Hi there!

[...]

PR #319, opened 2023-11-20
Fix/cache operation

drupol:
This PR:

- [x] Relates to https://github.com/loophp/collection/issues/264#issuecomment-1808384744
- [x] Fix the `cache` operation
- [ ] Provide ...
- [ ] Has documentation

cc @rela589n

### DIFF
--- a/docs/pages/code/operations/cache.php
+++ b/docs/pages/code/operations/cache.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App;
+
+use loophp\collection\Collection;
+
+include __DIR__ . '/../../../../vendor/autoload.php';
+
+$fopen = fopen(__DIR__ . '/vendor/autoload.php', 'rb');
+
+$collection = Collection::fromResource($fopen)
+    ->cache();
+
+for ($i = 0; 3 > $i; ++$i) {
+    foreach ($collection as $value) {
+        print_r($value);
+    }
+}
+
+$callable = static function () {
+    var_dump('Generator call');
+
+    yield 'a' => 'a';
+};
+
+$collection = Collection::fromCallable($callable)
+    ->cache();
+
+for ($i = 0; 3 > $i; ++$i) {
+    foreach ($collection as $value) {
+        print_r($value);
+    }
+}
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -89,7 +89,7 @@ final class Collection implements CollectionInterface, JsonSerializable, Countab
 
     public function cache(?CacheItemPoolInterface $cache = null): CollectionInterface
     {
-        return new self((new Operation\Cache())()($cache ?? new ArrayAdapter()), [$this]);
+        return new self((new Operation\Cache())()($cache ?? new ArrayAdapter()), [$this->squash()]);
     }
 
     public function chunk(int ...$sizes): CollectionInterface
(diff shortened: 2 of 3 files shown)


drupol:
Good for you @AlexandruGG ?

AlexandruGG:
> Good for you @AlexandruGG ?

Will review in a bit!

docs/pages/code/operations/cache.php
```diff
@@ -0,0 +1,35 @@
...
+
+$collection = Collection::fromResource($fopen)
+    ->cache();
+
+for ($i = 0; 3 > $i; ++$i) {
+    foreach ($collection as $value) {
+        print_r($value);
+    }
+}
+
```
AlexandruGG, line 21: can we pls add some comments to show what will be printed in each of these cases?

src/Collection.php
```diff
@@ -89,7 +89,7 @@ public function averages(): CollectionInterface
 
     public function cache(?CacheItemPoolInterface $cache = null): CollectionInterface
     {
-        return new self((new Operation\Cache())()($cache ?? new ArrayAdapter()), [$this]);
+        return new self((new Operation\Cache())()($cache ?? new ArrayAdapter()), [$this->squash()]);
```
AlexandruGG, line 92: to make sure I got this right: after this change calling `cache` will always eagerly evaluate all the previous operations?

If that's the case, I'm not sure this is really a "fix" but a change in functionality. It's fine to introduce it but I think it might be good to have it based on a boolean parameter and not enabled by default so as to not change the previous behviour. WDYT?

AlexandruGG (commented):
Do we have tests covering this?

docs/pages/code/operations/cache.php
drupol: Yes, will do

src/Collection.php
drupol: That's the whole point of this operation. Using `cache` should do what it says.

In the [following example](https://github.com/loophp/collection/issues/264#issuecomment-1808384744), the `cache` operation doesn't do at all what it is supposed to do, this PR fix this.

And you're totally right when you say that this is eagerly evaluating the Collection. I don't see we could do that differently at the moment. I wish there would be a better option to avoid evaluating eagerly, but I couldn't find anything better yet.

AlexandruGG: I think an important distinction might be needed here, but correct me if I'm wrong: in that example, `cache + all` is not doing what the user expects. Is this is a general issue with `cache`, is it working in any other cases?

If it is working in other cases then the behaviour change in this PR is breaking BC, which is why I was suggesting to only add `squash` based on a param, eg:

```php
public function cache(?CacheItemPoolInterface $cache = null, boolean $shouldCache = false): CollectionInterface
    {
        $withSquash = $shouldCache ? [$this->squash()] : [];
        return new self((new Operation\Cache())()($cache ?? new ArrayAdapter()), $withSquash);
    }
```

drupol: I'll check when I'm back in front of my laptop, away from computer right now :S

AlexandruGG: Hey @drupol, I assume you haven't had the chance to look at this yet? Please feel free to reopen when you do :)

drupol:
Indeed, not yet :( I definitely think that this operation should be reworked totally and the parameter removed.

I'll give it a shoot asap !

PR #322, opened 2023-11-24
feat: update `Collection::fromFile` constructor

drupol:
I saw a presentation made by a French company: https://speakerdeck.com/goathier/collection

In their slides, I had the feeling that parsing a CSV file was too complicated, so I had the idea to implement the feature in `loophp/collection`. But to do that, I had to update `loophp/iterators` first, see the relevant [commit](https://github.com/loophp/iterators/commit/46419a0213b2791727c94c894ab08e172e2fc88b).

This PR:

- [x] Update `Collection::fromFile` constructor in a non-breaking backward compatibility
- [x] It breaks backward compatibility (Minor BC - the signature of `Collection::fromFile()` has been updated, the `length` parameter has been replaced with the `consumer` parameter.)
- [x] Has documentation


alexandre-daubois:
Gauthier told me you created this PR, that's a big yes to me! This will definitely improve the DX when dealing with files. We'll definitely use this feature in production as soon as it is available 🚀

src/Collection.php
```diff
@@ -353,14 +353,14 @@ public static function fromCallable(callable $callable, iterable $parameters = [
     }
 
     /**
-     * @param null|non-negative-int $length
+     * @param Closure(resource): mixed $consumer
      *
-     * @return CollectionInterface<int, string>&self<int, string>
+     * @return CollectionInterface<int, string|mixed>&self<int, string|mixed>
      */
-    public static function fromFile(string $filepath, ?int $length = 2): CollectionInterface
```
AlexandruGG, line 360: hmm we could keep the `$length` parameter as well with a default value, I'm not really sure why should it be removed? Then the change is not breaking BC at all :).

And when you don't want to use it you can still call the constructor using the named argument:

```php
Collection::fromFile('data.csv', consumer: fgetcsv(...));
```

drupol: I decided to remove it since it can be put back through the `$consumer` as such:

```php
Collection::fromFile('data.csv', static fn ($resource) => fgets($resource, 3))
```

Therefore, it's not needed any more since we can now customize the `$consumer`.

AlexandruGG: fair enough, I guess you have more control overall now, let's go with it 👍

drupol: And the thing is that I removed it from `loophp/iterators` already... https://github.com/loophp/iterators/blob/main/src/ResourceIteratorAggregate.php

AlexandruGG (approved):

PR #340, opened 2024-04-02
Adds tap() - allows shorter syntax for chaining no side effects operations

ddebowczyk:
More convenient than `apply()` for chaining no side effect operations like logging or event dispatching.
Allows for shorter notation:

```php
$collection->tap(fn($item) => Event::dispatch(new ItemReceived($item)))
```

versus

```php
$collection->apply(function($item) {
   Event::dispatch(new ItemReceived($item));
   return true;
})
```

This PR:
- [+] Provides shorter notation for no side effect operations
- [-] It breaks backward compatibility
- [+] Is covered by unit tests
- [+] Has static analysis tests (psalm, phpstan)
- [+] Has documentation
- [-] Is an experimental thing

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -2287,6 +2287,22 @@ Signature: ``Collection::takeWhile(callable ...$callbacks): Collection;``
     Collection::fromIterable([1,2,3,4,5,6,7,8,9,1,2,3])
         ->takeWhile($isSmallerThanThree); // [1,2]
 
+tap
+~~~
+
+Execute callback(s) on each element of the collection.
+
+Iterates on the collection items regardless of the return value of the callback.
+
+It is a simplified version of the ``apply`` operation, as it does not require the callback to return a boolean value, allowing shorter syntax e.g. for logging or event dispatching: ``->tap(fn($value) => $events->emit(new ItemReceived($value)))``.
+
+Interface: `Tappable`_
+
+Signature: ``Collection::tap(callable ...$callbacks): Collection;``
+
+.. literalinclude:: code/operations/tap.php
+  :language: php
+
 transpose
 ~~~~~~~~~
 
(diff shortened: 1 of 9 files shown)


drupol:
Hi,

Thanks for the PR, it LGTM ! I wasn't aware that `apply` could be a burden to use and I don't mind adding `tap` into it. I agree that the syntax is simpler.

However, let's make sure tests are passing ;)

Let me know if you need some help.

ddebowczyk:
Frankly, i just copied tests from apply and removed irrelevant fragment
relying on boolean results. Not sure why they don't pass - any idea?

W dniu wt., 2.04.2024 o 08:06 Pol Dellaiera ***@***.***>
napisał(a):

drupol:
There's a fundemental difference between `apply` and `tap`. Of course they can do the same things, but since `apply` must returns a boolean, it can have a different way of working than `tap`... I guess the tests you've implemented are not reflecting that.

Also, it seems that there's an issue with the syntax.

To fix the codestyle:

```
./vendor/bin/grumphp run --tasks=phpcsfixer
``` 

To run the unit tests:

```
./vendor/bin/grumphp run --tasks=phpunit
``` 

To run the static analysis tests:

```
./vendor/bin/grumphp run --tasks=psalm.phpstan
```

ddebowczyk:
I cannot install pcov on my machine currently, so my `composer install` fails = I can't execute code checks, tests, etc on my local env. For the time being I will continue using my fork. I will try getting back to it when I find some time to set everything up properly. Thanks for quick replies.

drupol:
I understand. I wish I could help more concretely but right now, I'm busy on some other projects sadly.

If I can give a piece of advice to have an environment ready to use to develop in `loophp/collection` is to install `nix`:

```
curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install
```

To make it short, Nix is a universal package manager working on any Linux and Mac.

Once it is installed, got in the `loophp/collection` directory and do: 

```
nix shell github:loophp/nix-shell#env-php81 --impure
```

This will take 2 to 3 minutes. Once it's done, you'll have PHP 8.1 running with PCOV and everything you need to contribute to `loophp/collection` locally.

ddebowczyk:
Thanks again for the suggestion. I will get back with the fixes when I get tests running locally.

drupol:
That was fast! Did you used Nix in the end?

ddebowczyk:
Yes, Nix did the job. Thanks for suggesting it, saved me probably hours of install/config time.

drupol:
Excellent, hope you'll use it for some other stuff too !!!

drupol (approved):
LGTM, I'll merge it as it is and maybe do some optimisations later on.

Thank you very much for your contribution!

drupol:
This feature is available in the newly released version `7.6.0`.

Issue #350, opened 2024-11-26
More specific Chunkable::chunk() return type


rela589n:
Currently `Chunkable` interface looks like this:

```php
/**
 * @template TKey
 * @template T
 */
interface Chunkable
{
    /**
     * Chunk a collection of items into chunks of items of a given size.
     *
     * @see https://loophp-collection.readthedocs.io/en/stable/pages/api.html#chunk
     *
     * @return Collection<int, list<T>>
     */
    public function chunk(int ...$sizes): Collection;
}
```

As you can see, this operation returns `Collection<int, list<T>>`, which is fair. 

Yet, since after applying `chunk()`, each produced chunk can will never be empty list.

Therefore, maybe it would make sense to specify `Collection<int, non-empty-list<T>>` as a return type, since this is more specific.

drupol:
Hello,

Yes! Nice catch.

Please submit a PR to fix this and I'll merge it.

rela589n:
Closing as completed

PR #351, opened 2024-11-26
chore: more specific return type for chunk() method

rela589n:
Follows from https://github.com/loophp/collection/issues/350

### DIFF
--- a/src/Operation/Chunk.php
+++ b/src/Operation/Chunk.php
@@ -44,19 +44,15 @@ final class Chunk extends AbstractOperation
                             return;
                         }
 
-                        if (count($chunk) !== $size) {
-                            $chunk[] = $value;
+                        $chunk[] = $value;
 
-                            continue;
+                        if (count($chunk) >= $size) {
+                            ++$chunkIndex;
+
+                            yield $chunk;
+
+                            $chunk = [];
                         }
-
-                        ++$chunkIndex;
-
-                        /** @psalm-var non-empty-list $chunk */
-
-                        yield $chunk;
-
-                        $chunk = [$value];
                     }
 
                     if ([] !== $chunk) {
--- a/src/Operation/Chunk.php
+++ b/src/Operation/Chunk.php
@@ -18,47 +18,49 @@ use function count;
 final class Chunk extends AbstractOperation
 {
     /**
-     * @return Closure(int...): Closure(iterable<TKey, T>): Generator<int, list<T>>
+     * @return Closure(int...): Closure(iterable<TKey, T>): Generator<int, non-empty-list<T>>
      */
     public function __invoke(): Closure
     {
         return
             /**
-             * @return Closure(iterable<TKey, T>): Generator<int, list<T>>
+             * @return Closure(iterable<TKey, T>): Generator<int, non-empty-list<T>>
              */
             static fn (int ...$sizes): Closure =>
                 /**
                  * @param iterable<TKey, T> $iterable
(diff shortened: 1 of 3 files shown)


drupol:
I think there are some new errors to fix.

rela589n:
Do you mean this CI error?

image

Unfortunately I could not get my local environment ready for the development (If I remember correctly, if I try to run tests on the main branch, some of them fail) yet

src/Operation/Chunk.php
```diff
@@ -18,47 +18,49 @@
...
                  *
-                 * @return Generator<int, list<T>>
+                 * @return Generator<int, non-empty-list<T>>
                  */
                 static function (iterable $iterable) use ($sizes): Generator {
                     $sizesCount = count($sizes);
-                    $i = 0;
-                    $values = [];
+                    $chunkIndex = 0;
+                    $chunk = [];
```
rela589n, line 38: @drupol , I have slightly changed the variables names here, since it took me around 15 minutes to figure out what `$i`, `$values`, and `$sizes` variables mean and what the code below does. If you wish, I could revert it back to the previous version, but in my opinion the current implementation is better

src/Operation/Chunk.php
```diff
@@ -18,47 +18,49 @@
...
+                            $chunk[] = $value;
 
                             continue;
                         }
 
-                        ++$i;
+                        ++$chunkIndex;
 
-                        yield $values;
+                        /** @var non-empty-list<T> $chunk */
```
rela589n, line 56: without this `@var` annotation psalm won't be able to infer the fact that chunk is never empty when it is yielded

src/Operation/Chunk.php
```diff
@@ -18,47 +18,49 @@
 final class Chunk extends AbstractOperation
```
rela589n, line 18: @drupol , frankly speaking I have no clue what these errors are all about

image

src/Operation/Chunk.php
drupol: No no it's fine, it's actually a good idea.

rela589n:
@drupol , could you help me with this, since right now I'm not sure what I'm expected to do in order to make this PR merged?

rela589n:
@drupol , could you help me with this, since right now I'm not sure what I'm expected to do in order to get this PR merged?

drupol:
Sorry for the long delay, I'm quite busy with this end of year and I don't have much time left for my open-source projects sadly.

Here, it looks like there are code style stuff to update.
Your changes are also introducing new static analysis issues.

Are you able to run the tests on your local machine so you can fix them locally and commit/push when it's done?

src/Operation/Chunk.php
rela589n: Actually this didn't help either, because psalm treats `T` somewhat awkwardly.

Threrefore, I have somewhat rewritten the implementation into following variant so that both psalm and phpstan are satisfied:

```php
$chunk[] = $value;

if (count($chunk) >= $size) {
    ++$chunkIndex;

    yield $chunk;

    $chunk = [];
}
```

rela589n:
> I'm quite busy with this end of year and I don't have much time

It's all right, take your time

> Here, it looks like there are code style stuff to update.

I have updated the code so that all checks have passed now.
Once you have some bit of time, please review it.

Thank you,
Yevhen

drupol (approved):
LGTM! Thanks!

